Issue #11, opened 2022-10-17
Create a container image and build it


Angelmmiguel:
We want to simplify the usage of `wws` as much as we can. Since containers are quire popular these days, we want to provide a related `wasm-workers-server` container. This container will include the Rust binary and the minimum extra packages required.

We will compile this container for the following platforms:

**Required**
- amd64 / x86
- arm64

**Optional**
- armv7

The container will use the `/app` folder as the entrypoint for the diffrent worker files. So, the goal is that the following command will spawn the server directly:

```
podman run --rm --name wws -v ./my-workers:/app projects.registry.vmware.com/wasmlabs/containers/wasm-workers-server 
```

PR #15, opened 2022-10-19
Build images

assambar:
- Added code to build a multi-arch image
- Updated README and docs with info about said image (already published)


README.md
```diff
@@ -73,6 +73,20 @@ curl https://raw.githubusercontent.com/vmware-labs/wasm-workers-server/main/inst
   ./wws --help
 ```
 
+### Running in a container
+
+If you don't want to install anything locally you could just run `wws` from the `projects.registry.vmware.com/wasmlabs/containers/wasm-workers-server:latest` container image. All you need to do is:
```
ereslibre, line 78:
```suggestion
If you don't want to install anything locally you can just run `wws` from the `projects.registry.vmware.com/wasmlabs/containers/wasm-workers-server:latest` container image. All you need to do is:
```

ereslibre (approved):
LGTM, thanks @assambar! Just a minor wording suggestion.

README.md
assambar: Addressed. But force pushed.

docs/docs/containers.md
```diff
@@ -0,0 +1,26 @@
+---
+sidebar_position: 3
+---
+
+# Running in a container
+
+For convenience we have published a docker image that contains Wasm Workers Server. It is available at `projects.registry.vmware.com/wasmlabs/containers/wasm-workers-server:latest`. Any container that runs it will get the `wws` binary, running and:
```
Angelmmiguel, line 7: I would use here a "container image" as a generic term:

```suggestion
For convenience we have published a container image that contains Wasm Workers Server. It is available at `projects.registry.vmware.com/wasmlabs/containers/wasm-workers-server:latest`. Any container that runs it will get the `wws` binary, running and:
```

docs/docs/containers.md
```diff
@@ -0,0 +1,26 @@
...
+
+ - looking for worker handlers in the `/app` folder
+ - listening on `0.0.0.0:8080` inside the container
+
+The image is based on `debian:bullseye-slim` + the `wws` binary. It includes support for the `linux/amd64` and `linux/arm64/v8` platforms. The image size should be around `~100MiB`
```
Angelmmiguel, line 12: In the future, I would like to optimize for a minimal image. It seems to me that `~100MiB` for a single Rust binary may be too much. Anyway, something we can improve later :)

Angelmmiguel (approved):
I just added some minor comments @assambar! The PR looks great. I already approved it, so we can squash and merge after those.

Thanks!

docs/docs/containers.md
assambar: > In the future, I would like to optimize for a minimal image. It seems to me that `~100MiB` for a single Rust binary may be too much. Anyway, something we can improve later :)

I totally agree on the size. Will need to experiment if we can run wws in alpine out of the box and if not build it from there. Will give it a quick try and if it works I will updated the PR and re-publish the images.

assambar: @Angelmmiguel as expected the `wws` binary would not run out of the box on alpine, so I will do nothing about this for now. Let's log an issue for this.

docs/docs/containers.md
assambar: Addressed.

assambar:
https://github.com/vmware-labs/wasm-workers-server/issues/11

PR #18, opened 2022-10-23
Fix doc, Tutorials / Create a Rust worker

taisukef:
Thank you for the exciting project!


gzurl:
Thanks @taisukef for your patch! LGTM!

PR #19, opened 2022-10-23
add support IPv6 hostname

taisukef:
on previous version
```bash
wws --host "[::]"
```

```
Error: Custom { kind: Uncategorized, error: "failed to lookup address information: nodename nor servname provided, or not known" }
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -174,7 +174,7 @@ async fn main() -> std::io::Result<()> {
 
         app
     })
-    .bind((args.hostname.as_str(), args.port))?;
+    .bind(format!("{}:{}", args.hostname.as_str(), args.port))?;
 
     println!(
         "🚀 Start serving requests at http://{}:{}\n",


ereslibre:
Thank you @taisukef. I cannot reproduce this issue neither at main or with the 0.5.0 released version.

Angelmmiguel:
Hello @taisukef,

I checked the issue with @ereslibre and we were able to reproduce the issue. The current version works fine on Linux, but we get the same error you are pointing out in macOS. I also found a similar a related issue in `actix-web`, the framework we're using to manage the HTTP requests (https://github.com/actix/actix-web/issues/1320). 

The proposed solution in that issue looks similar to your PR. For us, it seems the `bind` method should be the one dealing with this case. Let us revisit the `bind` definition at `actix-web` and we will sync on the PR with you for the final solution.

Meanwhile, you can use the "extended" IPv6 for localhost:

```
wws --host "0:0:0:0:0:0:0:1"
```

Thank you for your PR!

ereslibre (approved):
As @Angelmmiguel said we could reproduce this both on Mac OS X and GNU/Linux. Only that it was taking a while to bind so I could not see the error after certain time had passed.

Thanks for the contribution @taisukef! LGTM

Angelmmiguel (approved):
Approved too! Thanks for your contribution :)

Issue #17, opened 2022-10-21
Cannot build examples/rust-basic 


NewGyu:
Sorry for the boring point, when running cargo build, the errors below was occurred.

```
$ cargo build --target wasm32-wasi --release
   Compiling rust-basic v0.1.0 (/workspaces/wasm-workers-server/examples/rust-basic)
error: invalid format string: expected `'}'`, found `'m'`
  --> src/main.rs:19:16
   |
19 |         body { max-width: 1000px; }
   |              - ^ expected `}` in format string
   |              |
   |              because of this opening brace
   |
   = note: if you intended to print `{`, you can escape it using `{{`
```

I think `{` and `}` should be replaced to `{{` and `}}`.

```
    <style>
        body {{ max-width: 1000px; }}
        main {{ margin: 5rem 0; }}
        h1, p {{ text-align: center; }}
        h1 {{ margin-bottom: 2rem; }}
        pre {{ font-size: .9rem; }}
        pre > code {{ padding: 2rem; }}
        p {{ margin-top: 2rem; }}
    </style>
```

Thank you.

PR #20, opened 2022-10-23
fix example, rust-basic, add escape

taisukef:
Fixes: https://github.com/vmware-labs/wasm-workers-server/issues/17


ereslibre (approved):
LGTM, thank you @taisukef!

PR #23, opened 2022-11-10
feat: add new workflow to cross-compile wws

### LINKED ISSUE #6: Add automation to build the different binaries in every release
For every release, we should build and upload the different binaries. With this automation we ensure people see where those files come from and how they were built. We will upload the binaries and also the SHA256 of every of them. We will print these values in the job too.

For now, we want to compile it for these different platforms:

**Required**

- (macOS) Darwin x86
- (macOS) Darwin ARM64
- Linux x86

**Optional**

- Linux ARM64
- Windows 64
- Raspberry Pi ARM8 
- RISC-V (if possible)

Angelmmiguel:
Create a new GitHub Action workflow to cross-compile Wasm Workers Server to:

- Linux
- Windows
- Mac

In both architectures:

- `x86_64`
- `aarch64`

For the `x86_64` builds, I'm using the machines provided by GitHub. In the case of `aarch64` builds, I'm using the [cross](https://github.com/cross-rs/cross) tool. This is required as GitHub is not providing `aarch64` machines.

The end goal is that the project runs this workflow to build the binaries on every new tag. The release will be crafted manually by the team to ensure release notes are correct.

It closes #6

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -106,5 +106,3 @@ jobs:
       with:
         name: wws-${{ matrix.name }}-${{ matrix.arch }}.tar.gz
         path: ${{ env.TARBALL }}
-        # TODO: This is for testing
-        retention-days: 7
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -1,13 +1,13 @@
-name: Rust release
+name: Build artifacts
 
 on:
   push:
     branches: [ main ]
     tags:
     - "[0-9]+.[0-9]+.[0-9]+"
-  # TODO: For tesgting. Run this workflow only on a new release.
-  pull_request:
-    branches: [ main ]
+  # TODO: uncomment for testing. This will run by default only when a new tag is created
+  # pull_request:
+  #   branches: [ main ]
 
 jobs:
   test:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -97,12 +97,12 @@ jobs:
           cp "target/${{ matrix.arch }}-${{ matrix.platform }}/release/wws" ./out
         fi
 
-        tar czf "wws-${{ matrix.name }}-${{ matrix.arch }}.tar.gz" ./out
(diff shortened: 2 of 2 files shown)


.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,108 @@
+name: Build artifacts
+
+on:
+  push:
+    branches: [ main ]
+    tags:
+    - "[0-9]+.[0-9]+.[0-9]+"
```
ereslibre, line 7: Is this regexp matching `vX.Y.Z` as we are tagging at the moment?

.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,108 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Caching
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
ereslibre, line 24: I would not use caching in conjunction with artifacts meant to be released. This is so we rule out all compiler/toolchain caching issues. Releasing should not happen that often (even if we take into account this is currently set up for push on `main` as a "latest" artifact).

assambar (approved):
Ship it!

.github/workflows/artifacts.yml
Angelmmiguel: Sounds good! I will apply this change directly on `main` as I already merged the changes 👍

.github/workflows/artifacts.yml
Angelmmiguel: You are right @ereslibre. I fixed it on 57c731592cca2a811614caa597135810bc1b2800

.github/workflows/artifacts.yml
Angelmmiguel: Fixed it on bc4c76ede2352c92868e280bf43989fc45fcaa71

PR #27, opened 2022-11-14
feat: upgrade wasmtime to 2.0.2

### LINKED ISSUE #26: Upgrade wasmtime to 2.0.2
Upgrade the proejct to use the latest features and fixes from the Wasmtime project: https://github.com/bytecodealliance/wasmtime/releases/tag/v2.0.2.

Angelmmiguel:
Upgrade wasmtime to the latest version. I will wait for #23 to be merged and publish a new version of the project.

It closes #26

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,10 +10,10 @@ name = "wws"
 path = "src/main.rs"
 
 [dependencies]
-wasmtime = "2.0.0"
-wasmtime-wasi = "2.0.0"
+wasmtime = "2.0.2"
+wasmtime-wasi = "2.0.2"
 anyhow = "1.0.66"
-wasi-common = "2.0.0"
+wasi-common = "2.0.2"
 actix-web = "4"
 env_logger = "0.9.0"
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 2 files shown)


gzurl:
LGTM!

gzurl (approved):

Cargo.toml
```diff
@@ -10,10 +10,10 @@ name = "wws"
 path = "src/main.rs"
 
 [dependencies]
-wasmtime = "2.0.0"
-wasmtime-wasi = "2.0.0"
+wasmtime = "2.0.2"
```
ereslibre, line 13: We could keep less strict dependencies in Cargo.toml here, such as `wasmtime = "2.0"`.

This applies to all deps that are specified as major.minor.patch in dependencies. I think pinning the whole dep makes sense in certain circumstances (when a specific patch level is desired for a specific reason), but in general I think we want to move forward the whole dependency tree when we do a cargo update.

ereslibre (commented):
LGTM; thanks. A minor comment. We can split the commits if we don't want to change all requirements together.

Cargo.toml
Angelmmiguel: That's a good point. I would revisit this in a separate task to apply a common policy for all dependencies. For example, to pin them to `major.minor` and keep the patch to be updated automatically.

Issue #3, opened 2022-10-10
Improve path management in wws


Angelmmiguel:
The `wws` server relies on the local filesystem heavily. The way we are managing paths at this point may cause issues when running the server in Windows environments. In addition to that, the way we check for multiple extensions is quite complex and error prone.

The goal for this issue is to:

- Define available extensions in a constant (in the future we may mek this configurable)
- Limit the conversion of paths to `String` and `&str` to printing data
- Rely on `Path` and `PathBuf` for any path management requirement

Angelmmiguel:
@assambar thanks for the report! The issue will be fixed in #30 😄

Angelmmiguel:
The main issue we have here is the limitations of the [`glob` library](https://docs.rs/glob/0.3.0/glob/). The library doesn't support patterns with multiple extensions like `**/*.{js,wasm}`. We can chain multiple paths (iterators), but if the extensions are not hardcoded, [we will need to `flatten`](https://stackoverflow.com/a/60371634) all the paths while we're reading the files. 

The [`wax` library](https://github.com/olson-sean-k/wax) is more up-to-date and covers the different use cases. I will switch to use this one.

PR #29, opened 2022-11-14
feat: read static files from the public folder

### LINKED ISSUE #7: Support static assests in handlers
Some handlers return HTML code that may require static assets such as JavaScript (client-side) and CSS. Currently, it can be injected in the handler, however is much more convenient to provide a way to set static files.

We can provide this feature using two approaches:

* Mount generic folders (using WASI) in the handler and provide libraries to manage those files
* Add a new `data.folder` configuration property that will serve the files directly using the Rust server

The final approach is still under discussion.

Angelmmiguel:
Support static files by reading them from the `public` folder in the root of the project. Any file inside this folder will be served as a static file, so project like static sites can work without embedding files in the workers. This is a pretty consistent standard in different web frameworks like:

* [NextJS](https://nextjs.org/docs/basic-features/static-file-serving)
* [Astro](https://docs.astro.build/en/core-concepts/project-structure/#public)
* [Remix](https://remix.run/docs/en/v1/tutorials/jokes#explore-the-project-structure)
* etc

Note that the `public` folder based on the given path in the CLI. So, for `wws ./examples`, the CLI will serve the files from `./examples/public`. There's an issue with path management as `wws ./examples` should calculate the routes without including the `examples` folder. This issue will be solved in #3.

It closes #7


gzurl (approved):
LGTM!

Angelmmiguel:
Thanks for the review @gzurl ! I'm going to add the missing use case of allowing `html` files in the `public` folder.

assambar (approved):
Ship it!

You could add this later, but I think a simple example worker that uses static assets will help the adoption of this feature.
IDK, for example something simple that creates a business card with some background image based on a call to `make-business-card.js?first_name=John&last_name=Doe&phone=123345...`

Angelmmiguel:
@gzurl @assambar @ereslibre I completed the feature by adding support for static HTML files. Now, the `/public/about.html` file will be mounted in `/about` automatically.

> You could add this later, but I think a simple example worker that uses static assets will help the adoption of this feature.
IDK, for example something simple that creates a business card with some background image based on a call to make-business-card.js?first_name=John&last_name=Doe&phone=123345...

This is an interesting use case. It can work for OpenGraph images and business cards (as you suggested). I will give a second though to that idea and will implement some examples around. However, those will work purely with workers, so I don't think static assets will be used that much in the example.

src/main.rs
```diff
@@ -169,6 +192,51 @@ async fn main() -> std::io::Result<()> {
...
+                        Path::new(base_path).join(format!("public{}.html", uri_path));
+                    let public_404_path = Path::new(base_path).join("public").join("404.html");
+
+                    if uri_path.ends_with("/") && index_folder_path.exists() {
+                        file = NamedFile::open_async(index_folder_path).await;
+                    } else if !uri_path.ends_with("/") && html_ext_path.exists() {
+                        file = NamedFile::open_async(html_ext_path).await;
+                    } else {
+                        file = NamedFile::open_async(public_404_path).await;
+                    }
```
ereslibre, line 225: I think this logic could be extracted to a couple of functions, just to keep this closure a bit more readable.

ereslibre (approved):
Looking fantastic; thanks @Angelmmiguel! LGTM. A minor optional comment.

src/main.rs
Angelmmiguel: I extracted this logic into a separate method. Thanks for the suggestion!

PR #32, opened 2022-11-17
fix: normalize API routes using Rust path entities

### LINKED ISSUE #30: Fix wrong routes when passing a path as an argument
_From #3 and #29_

When passing a path to `wws` (like `wws ./examples`), it should load the files starting from that route and assign the HTTP endpoint from it too. Currently, this is not the case and running `wws ./examples` when `./examples/handler.js` exists will assign the `/examples/handler` route instead of the desired `/handler`. 

The issue is located at:

https://github.com/vmware-labs/wasm-workers-server/blob/4837e1b5bc9637b1a24245627f59c4567442fa60/src/router.rs#L71

Here we are removing `base_path` from `api_path`, but in this case the string doesn't match properly:

```
api_path: examples/handler.js
base_path: ./examples/
```

In addition to that, in [#3 @assambar](https://github.com/vmware-labs/wasm-workers-server/issues/3#issuecomment-1282582912) described an issue when the folder does not include the trailing `/`. The trailing slash should be optional, so we need to fix that issue too.

Angelmmiguel:
In the current `main`  version, file paths are not normalized properlies as APIs. This is causing inconsistent behaviors in different environments:

- In unix, the API path includes the root folder the CLI receives as argument. For example, for `wws ./examples/js-basic`, the URL of the `handler.js` should be `/handler`. Note that ` ./examples/js-basic` is the local folder to read the files, not a prefix for the final URL. However, the HTTP endpoint it generates is `/examples/js-basic/handler.
- In windows, it just doesn't work. The `handler.js` is not accessible (See [1]).

This is mainly caused by converting file paths to string and apply transformations later. Since paths may include special components such as `.`  and `./`, applying operations like `replace` can have different outcomes depending on the environment. This also causes inconsistent behaviors when a final backslash is added or not to the CLI argument.

In this PR, I changed that implementation to run all required transformations directly on the path. Instead of replacing and subtracting from a converted string, I iterate over the components to remove any special entity and keep only the [Normal ones](https://doc.rust-lang.org/std/path/enum.Component.html). Finally, I convert them to a String to be used as the API path.

I tested the current changes on both Windows and Mac (Unix) and they work great:

win_working

[1]
win_err
win_err2

It closes #30

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   test:
-    stretegy:
+    strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,11 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    stretegy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
     - name: Caching
@@ -23,23 +27,11 @@ jobs:
           target
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
     - name: Format
+      if: ${{ matrix.os == 'ubuntu-latest' }}
       run: cargo fmt
     - name: Clippy
+      if: ${{ matrix.os == 'ubuntu-latest' }}
       run: cargo clippy
-    - name: Test
-      run: cargo test
-  win-test:
-    runs-on: windows-latest
-    steps:
-    - uses: actions/checkout@v3
-    - name: Caching
-      uses: actions/cache@v3
-      with:
-        path: |
-          ~/.cargo/registry
-          ~/.cargo/git
-          target
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
     - name: Test
       run: cargo test
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,20 @@ jobs:
       run: cargo clippy
     - name: Test
       run: cargo test
+  win-test:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Caching
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+    - name: Test
+      run: cargo test
   build:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
           target
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
     - name: Format
-      run: cargo fmt
+      run: cargo fmt --check
     - name: Clippy
       run: cargo clippy
   test:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,22 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Caching
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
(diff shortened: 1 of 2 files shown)


Angelmmiguel:
I found an error when testing subfolders in Windows. It's using the `\` separator and that's causing actix not to define the route properly.

.github/workflows/build.yml
```diff
@@ -28,6 +28,20 @@ jobs:
...
+    - name: Caching
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+    - name: Test
+      run: cargo test
```
ereslibre, line 44: Both jobs can be replaced with a single one and a matrix strategy, such as:

```yaml
jobs:
  test:
    strategy:
      fail-fast: false
      matrix:
        os: [ubuntu-latest, windows-latest]
    runs-on: ${{ matrix.os }}
    steps:
    - name: Checkout
      uses: actions/checkout@v3
    - name: Format
      run: cargo fmt
    - name: Clippy
      run: cargo clippy
    - name: Test
      run: cargo test
```

src/router.rs
```diff
@@ -65,26 +66,53 @@ impl Route {
...
 
-            normalized
-        } else {
-            // TODO: Manage better unexpected characters in paths
-            String::from(path.to_str().unwrap_or("/unknown"))
+            if let Some(parsed_str) = os_str.to_str() {
+                // Force the separator to be / instead of \ in Windows
+                normalized_path.push('/');
+                normalized_path.push_str(parsed_str);
+            }
```
ereslibre, line 112: Can this `filter` be replaced with `filter_map` and do the mapping and filtering at once? Is it more clear that way?

src/router.rs
```diff
@@ -65,26 +66,53 @@ impl Route {
...
+        let mut normalized_path = String::new();
+
+        for c in clean_comps {
+            let os_str = c.as_os_str();
 
-            normalized
-        } else {
-            // TODO: Manage better unexpected characters in paths
-            String::from(path.to_str().unwrap_or("/unknown"))
+            if let Some(parsed_str) = os_str.to_str() {
```
ereslibre, line 108: I fear this on Windows will map a path like "some\path\to\project" into "/some\path\to\project", because although we are forcing a root '/', the rest of separators produced by 'os_str.to_str()' under Windows will contain '\' instead.

ereslibre (commented):
Looking great, lots of tests! A couple of comments :)

src/router.rs
Angelmmiguel: At this point, the separator should be ignored. Note that `c` represents a `Component` which doesn't include the platform separator: 

```
# pseudocode
Components("a\b\c") == Components("a/b/c")
```

That was the trick on this PR. Instead of managing entire path strings, I converted them to components and stripped non "Normal" ones.

src/router.rs
Angelmmiguel: Great suggestion!! I think I got it 😄

.github/workflows/build.yml
Angelmmiguel: Sounds good! I wasn't running `clippy` nor `fmt` on Windows, but I can easily skip those using conditionals.

.github/workflows/build.yml
```diff
@@ -23,8 +27,10 @@ jobs:
           target
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
     - name: Format
+      if: ${{ matrix.os == 'ubuntu-latest' }}
       run: cargo fmt
     - name: Clippy
+      if: ${{ matrix.os == 'ubuntu-latest' }}
```
ereslibre, line 33: This is probably a matter of personal taste. I find it more clear to specify two jobs (one for linting -- fmt + clippy), and another job for test that contains the matrix.

Skipping 2 steps in a matrix that is NxM so that it skips these steps in all instances except in one is going to skip a lot, specially when the matrix grows. Also, jobs run in parallel whereas steps run in sequence.

ereslibre (approved):
LGTM, thanks @Angelmmiguel! It looks great! Just a small optional comment about the GH workflow setup

src/router.rs
ereslibre: Right, thanks!

.github/workflows/build.yml
```diff
@@ -23,8 +27,10 @@ jobs:
           target
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
     - name: Format
+      if: ${{ matrix.os == 'ubuntu-latest' }}
       run: cargo fmt
```
ereslibre, line 31:
```suggestion
      run: cargo fmt --check
```

Angelmmiguel: Applied 👍

ereslibre (approved):
🎉

assambar (approved):
Ship it!

PR #37, opened 2022-11-21
feat: add the prefix option to prepend a path to all URLs

### LINKED ISSUE #31: Add the prefix option to the CLI
The `--prefix` option allows users to set a prefix for all the routes. This is useful when a specific application is mounted in a subpath of a server. Currently, this is reproduced by using a folder `wws ./examples`, but this will change after we fix #30.

The `--prefix` will be optional and expect a `String` as type.

Angelmmiguel:
Introduce a new `--prefix` option to preped a string to all URLs (workers and static assets). This option receives a string that will be properly normalized by the application. If no option is provided, the default value will be "" (no prefix).

This is the outcome from the `--help` command:

```
~/W/o/webassembly-functionless ❯❯❯ cargo run -- --help

Wasm Workers Server is a blazing-fast self-contained server that routes HTTP requests to workers in your filesystem. Everything run in a WebAssembly sandbox

Usage: wws [OPTIONS] [PATH]

Arguments:
  [PATH]  Folder to read WebAssembly modules from [default: .]

Options:
      --host <HOSTNAME>  Hostname to initiate the server [default: 127.0.0.1]
  -p, --port <PORT>      Port to initiate the server [default: 8080]
      --prefix <PREFIX>  Prepend the given path to all URLs
  -h, --help             Print help information
  -V, --version          Print version information
```

Note that this string must use URI format (`/X/Y`) because it's not a filesystem path. However, I also considered the case that someone introduces a windows path (`\X\Y`). I considered it because it's easy to implement, but I may drop support in the future.

It closes #31

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -149,8 +149,8 @@ pub fn initialize_routes(base_path: &Path, prefix: &str) -> Vec<Route> {
 /// \app. This shouldn't be considered as "prefix" must be an URI
 /// path. However, the check is pretty simple, so we will consider
 /// it.
-pub fn format_prefix(source: Option<String>) -> String {
-    let mut normalized_prefix = source.unwrap_or_default();
+pub fn format_prefix(source: &str) -> String {
+    let mut normalized_prefix = source.to_string();
     // Ensure the prefix doesn't include any \ character
     normalized_prefix = normalized_prefix.replace('\\', "/");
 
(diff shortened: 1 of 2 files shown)


src/router.rs
```diff
@@ -137,6 +138,35 @@ pub fn initialize_routes(base_path: &Path) -> Vec<Route> {
...
+/// To be flexible, the method will manage "windows" paths too:
+/// \app. This shouldn't be considered as "prefix" must be an URI
+/// path. However, the check is pretty simple, so we will consider
+/// it.
+pub fn format_prefix(source: &str) -> String {
+    let mut normalized_prefix = source.to_string();
+    // Ensure the prefix doesn't include any \ character
+    normalized_prefix = normalized_prefix.replace('\\', "/");
+
+    if !normalized_prefix.is_empty() {
```
assambar, line 157: Not sure if this is the rustacean way to go, but an early return `if source.is_empty()` in the very beginning of the function will save this if here as well as calling replace above, when it's not necessary

assambar (approved):
Ship it!

src/router.rs
Angelmmiguel: I was thinking the same. For me, it seems more natural to have early returns (ruby style), so I'll apply your suggestion.

PR #38, opened 2022-11-21
fix: avoid service JS files inside public folder as worker

### LINKED ISSUE #36: Skip any file in the public folder when loading routes
In #7, we added support for static assets using a special folder called `public`. This is causing an unexpected behavior when `public` contains static `.js` files for the UI. `wws` is loading those as `workers` when it should be considered static.

When loading routes in `initialize_routes`, we should ignore any file starting with `public`.

Angelmmiguel:
Ensure that assets in the `public` folder are always served as static. Before this PR, `.js` files inside the `public` folder will be processed as workers.

It closes #36

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -422,7 +422,7 @@ mod tests {
             assert_eq!(is_in_public_folder(&Path::new(t.0)), t.1)
         }
     }
-    
+
     #[test]
     fn format_provided_prefix() {
         let tests = [
--- a/src/router.rs
+++ b/src/router.rs
@@ -127,14 +127,24 @@ pub fn initialize_routes(base_path: &Path) -> Vec<Route> {
 
     for entry in glob_items {
         match entry {
-            Ok(filepath) => {
+            // Avoid loading static assets
+            Ok(filepath) if !is_in_public_folder(&filepath) => {
                 routes.push(Route::new(base_path, filepath));
             }
             Err(e) => println!("Could not read the file {:?}", e),
+            _ => {}
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/router.rs
```diff
@@ -128,16 +128,26 @@ pub fn initialize_routes(base_path: &Path, prefix: &str) -> Vec<Route> {
...
         }
     }
 
     routes
 }
 
+/// Checks if the given filepath is inside the "public" folder
+fn is_in_public_folder(path: &Path) -> bool {
+    path.components().any(|c| match c {
+        Component::Normal(os_str) => os_str == OsStr::new("public"),
```
ereslibre, line 146: I think as a first approach is fine, but a `public` named directory can exist deep on the project structure, and maybe meaning something else than a `public` toplevel for example, or `_public` etc...

I don't have an ideal strategy to propose right now, but we could add a note here on the code and possibly open an issue to keep track of that.

ereslibre (approved):
LGTM! A question regarding `public` named directories.

src/router.rs
Angelmmiguel: This is a good point @ereslibre. From a coding perspective, we can check it by joining `base_path` + `path`. However, I believe I'll keep current implementation for the initial release. I still want to understand from users if it's more convenient to support just a `/public` folder a the root (`base_path`) or if they prefer to have N `public` folders in the root or inside subfolders. 

The `public` folder is a pretty common pattern for static assets, so I don't expect people creating `public` folders for workers.

PR #42, opened 2022-11-23
feat: configure and inject environment variables in workers

### LINKED ISSUE #34: Add support for environment variables
Environment variables will be configured in `TOML` files. Every worker will define its own set of environment variable, which will be accessible only for them. 

As a tentative API, we will include a new `.env` resource in the TOML config file:

```
name = "test"
version = "1"

[env]
MY_ENV_VAR = "Testing!"
```

Then, the environment variables will be passed to the inner modules via WASI. The environment variables will be accessible via default env methods from the different languages.

Angelmmiguel:
Configure environment variables via configuration files (`TOML` files with the same name of the worker). These variables will be configured in the `vars` section. Teh value can be static (string) or inherited from the current context (`$VAR`). In the second case, `wws` will initialized it to an empty string if the given environment variable doesn't exist in the current context.

This is an example of configuration file:

```toml
name = "test"
version = "1"

[vars]
MESSAGE = "Hello 👋! This message comes from an environment variable"
```

The usage of environment variables depends on the selected language:

* **Rust**: retrieve them using `std::env`
* **JavaScript**: environment variables are injected as global constants

You have all the information and examples in the documentation I added to this PR.

It closes #34


docs/docs/tutorials/rust-workers.md
```diff
@@ -229,6 +229,46 @@ To add a KV store to your worker, follow these steps:
 
 1. Finally, open <http://127.0.0.1:8080/worker-kv> in your browser.
 
+## Read environment variables
+
+Environment variables are configured [via the related TOML configuration file](../features/environment-variables.md). These variables are directly injected as global constants in your worker. To read them, just use the same name you configured in your TOML file:
```
assambar, line 234:
```suggestion
Environment variables are configured [via the related TOML configuration file](../features/environment-variables.md). These variables are accessible via `std::env` in your worker. To read them, just use the same name you configured in your TOML file:
```

assambar (approved):
Greate feature! Just one small note on the docs.

src/runner.rs
```diff
@@ -160,12 +160,16 @@ impl Runner {
...
+        let mut wasi = WasiCtxBuilder::new()
             .stdin(Box::new(stdin))
             .stdout(Box::new(stdout.clone()))
             .stderr(Box::new(stderr))
             .inherit_args()?
             .build();
+        // Configure environment variables
+        for (key, value) in vars {
+            wasi.push_env(key, value)?;
+        }
```
ereslibre, line 172: We possibly can use `.envs()` from `WasiCtxBuilder` so we can set the envvars before calling to `.build()` instead of using `.push_env` (requiring mutability).

ereslibre (approved):
Really great contribution. LGTM, a minor comment to get rid of mutability on the WASI context builder.

src/runner.rs
Angelmmiguel: Good point! Let me try this solution 👍

Angelmmiguel: Done! I've just changed it to use `envs`

PR #45, opened 2022-11-28
feat: allow returning an array of bytes from Rust workers

### LINKED ISSUE #44: Allow Rust workers to return an array of bytes too
Rust workers return a `http::Response` object that it's lately serialized into a `UTF-8 JSON` string. Then, the JSON is returned to the main `wws` service via WASI `stdout`. Note that the current seralized JSON includes the body and any other related metadata such as headers and status code.

This beahvior limits the capabilities of the workers as some tasks may return non UTF-8 entities. For example, we may want to generate a binary file that can be represented with a `Vec<u8>` in Rust, but not as a `String`. For this reason, we plan to add support for an array of bytes (`Vec<u8>`) in Rust workers too. **This will introduce a breaking change** as the Rust kit assumes the worker will return a `String`. 

Supporting multiple values in Rust is simple, although we still need to make it serializable to JSON. To accomplish this, we plan to encode non-String values with `base64`. This is an initial iteration as `base64` will introduce a performance penalty. In the future, we will stop using JSON in favor of other formats that supports all the use cases without `base64`.

Angelmmiguel:
Allow Rust workers to return an array of bytes instead of a String. Note that `wws` serializes the worker output as an UTF-8 JSON, so an arbitrary array of bytes may not be serializable. For this reason, we implemented this feature using `base64` to encode the output when the worker exports an array of bytes.

Then, `wws` will look for the `base64` flag in the serialized JSON. If it's true, it will perform the decoding process before sending the result back.

The feature introduces a breaking change as we cannot now set the response body type as `String`. I created a new `Content` enum that can represents both text and base64 responses. 

The change is pretty trivial:

```rust
# Before
fn handler(_req: Request<String>, cache: &mut Cache) -> Result<Response<String>>;

# After
fn handler(_req: Request<String>, cache: &mut Cache) -> Result<Response<Content>>;
``` 

Thanks to the implementation of the `From` trait, the way you pass the body to the response will continue as it's now:

```rust
Ok(http::Response::builder()
    .status(200)
    .header("x-generated-by", "wasm-workers-server")
    .body(response.into())?)
```

The statement `response.into()` will initialize the right `Content` enum type.

The PR also updates the documentation and adds a new fun example called `pdf-create`.

It closes #44


examples/pdf-create/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+    current_layer.use_text(req.body(), 18.0, Mm(20.0), Mm(170.0), &font);
+    current_layer.use_text(
+        "Created by a Wasm module in Wasm Workers Server",
+        12.0,
+        Mm(20.0),
+        Mm(20.0),
+        &font,
+    );
+
+    doc.save(&mut buf).unwrap();
```
gzurl, line 28: Shouldn't this `Result<(), Error>` be handled in case the doc cannot be written?

gzurl:
LGTM! Just a minor comment.

src/runner.rs
```diff
@@ -47,16 +48,51 @@ impl WasmInput {
...
+        }
+    }
+
+    /// Return the content body as bytes. It will automatically
+    /// decode the data if the base64 flag is enabled.
+    pub fn body(&self) -> Result<Vec<u8>> {
+        if self.base64 {
+            Ok(decode(&self.data)?)
+        } else {
+            Ok(self.data.clone().into_bytes())
```
ereslibre, line 93:
```suggestion
            Ok(self.data.as_bytes().into())
```

Although we have to still clone when [transforming `&[u8]` to `Vec<u8>`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from-16) with the suggestion I propose, we are skipping a `String::clone()` that is going to be dropped immediately, as we are doing in the current code.

ereslibre (approved):
A new feature with more idiomatic code here and there, great work! 👏 🚢 

Minor comments!

examples/pdf-create/src/main.rs
Angelmmiguel: Good catch. I will bubble up the error by using the `?` operator :)

kits/rust/src/content.rs
```diff
@@ -0,0 +1,34 @@
...
+/// serialized UTF-8 JSONs. An array of bytes response may include
+/// bytes that cannot be represented as UTF-8. To avoid this
+/// limitation, Content is able to encode them as base64. Then,
+/// wws will ensure to decode them before sending the bytes to the
+/// client.
+
+// pub struct Content {
+//     pub data: String,
+//     pub base64: bool,
+// }
```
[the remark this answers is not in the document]
Angelmmiguel: 🤦

src/runner.rs
Angelmmiguel: Great catch! I'm gonna apply the change 👍

Issue #24, opened 2022-11-10
Optimize the size of the container image


assambar:
Currently the image is based on `debian:bullseye-slim` which is around 100MB. This is a huge overhead just for the sake of the wws binary.
We have to investigate if we can do an image based on something smaller, say alpine.

Angelmmiguel:
This issue was closed by #52 😄

PR #52, opened 2022-11-29
feat: automate container build with GitHub Actions

### LINKED ISSUE #50: Create the container image in GitHub Actions
`wws` is also provided as a container imager for convenience. However, this image is still generated manually. Since we plan to start using GitHub registry to publish them, we want to start using GitHub actions to publish it. These are the tags that should be created:

- A `preview` tag on every commin in `main`
- A `X.X.X` tag per new released version
- A `latest` tag pointing to the latest released version

Angelmmiguel:
Automate the creation of containers using GitHub Actions. This will push the images to the GitHub Package Registry. In addition to that, we are also trying to reduce the size of the produced images.

I created two different actions:

- Preview: runs on every commit in `main`
- Release: runs on every new vX.X.X tag

The code is duplicated, but I plan to refactor all actions and make some parts reusable in a separate PR.

Here you have the created package: https://github.com/vmware-labs/wasm-workers-server/pkgs/container/wws 

Closes #50 and #24

### DIFF
--- a/.github/workflows/container-preview.yml
+++ b/.github/workflows/container-preview.yml
@@ -1,0 +1,39 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "main" ]
+  # TODO: uncomment for testing. This will run by default only when a new commit is push to main
+  pull_request:
+    branches: [ "main" ]
+
+env:
+  CARGO_TERM_COLOR: always
+  REGISTRY: ghcr.io
+  IMAGE_NAME: vmware-labs/wws
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Repository clone
+      uses: actions/checkout@v3
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v2
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
+    - name: Login to Docker Hub
+      uses: docker/login-action@v2
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: image/Dockerfile
+          # TODO: First check the compilation and then
+          push: false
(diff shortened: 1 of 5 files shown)


.github/workflows/container-preview.yml
```diff
@@ -0,0 +1,96 @@
+# Copyright 2022 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+name: Container Build Preview
+
+on:
+  push:
+    branches: [ "main" ]
+  # TODO: uncomment for testing. This will run by default only when a new commit is push to main
+  # pull_request:
+  #   branches: [ "main" ]
```
ereslibre, line 10:
```suggestion
```

I would remove this comment here, I don't think we want to try to build the preview for PR's

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,97 @@
...
+# SPDX-License-Identifier: Apache-2.0
+name: Container Build Release
+
+on:
+  push:
+    tags:
+    - "v[0-9]+.[0-9]+.[0-9]+"
+  # TODO: uncomment for testing. This will run by default only when a new commit is push to main
+  # pull_request:
+  #   branches: [ "main" ]
```
ereslibre, line 11:
```suggestion
```

I would remove this comment. In order to try the workflow, I think is safer to fork on another place (like a personal account), and open the PR when everything is fine.

ereslibre (commented):
Great! 👏, some comments.

.github/workflows/container-preview.yml
```diff
@@ -0,0 +1,96 @@
...
+        include:
+          - arch: x86_64
+            cross: false
+            platform: unknown-linux-musl
+          - arch: aarch64
+            cross: true
+            platform: unknown-linux-musl
+    env:
+      # We use cross for arm builds
+      CARGO: cargo
```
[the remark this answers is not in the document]
Angelmmiguel: Good point! I updated it 😄

.github/workflows/container-preview.yml
Angelmmiguel: Makes sense. I'm gonna remove it 👍

ereslibre (approved):
🚢 it! 👏

PR #55, opened 2022-12-07
feat: allow dynamic routing based on file paths

### LINKED ISSUE #10: Allow dynamic parameters in URLs
The parameters will be provided based on the filename. As other projects using this approach do (like [NextJS](https://nextjs.org/docs/routing/dynamic-routes)), we will consider `[id].wasm` files as dynamic URLs with an `:id` parameter.

We also need to consider different scenarios. For example, a `/a/b` route may be replied by:

- `/a/b.js`
- `/a/[id].js`
- `/[id]/b.wasm`
- `/[id]/[other].wasm`

We need to establish a priority. The lower of the returned number, the more priority it has. This number is calculated based on the number of used parameters, as fixed routes has more priority than parameted ones.

To avoid collisions like `[id]/b.wasm` vs `/a/[id].js`. Every depth level will add an extra `+1` to the score. So, in case of `[id]/b.wasm` vs `/a/[id].js`, the /a/b path will be managed by `[id]/b.wasm`.

This parameter information will be injected to modules as part of the requests snapshots, allowing to create a new set of applications.

Angelmmiguel:
Introduce dynamic routes with parameters in files. For example, `[id].wasm` or `users/[id]/show`. The server identifies this pattern in files and configure actix to reply to parametrized URLs.

Note that the same `wasm_handler` serves all the requests. This means that `wws` calculates the best worker to reply on every request. The reasons I coded it in that way are:

- We fully control collisions when two workers can reply to the same request (see the #10 for more information). Actix will reply with the first handler that matches. However, users cannot set the handlers manually, so we want to ensure a consistent behavior.
- To map an actix path with a runner, we need to create factory service. Note that Actix will create an instance per thread (worker), so `Runner` instances cannot be shared. This will require multiple instances of the same Wasm module, so the resource consumption will be increased.

I updated the `Rust` and `JavaScript` kits and introduced some new examples so you can see how parameter in URLs works. 

It closes #10

## Pending

- [x] Improve the UI of the JS examples
- [x] Add a new Rust example
- [x] Update the documentation


src/router.rs
```diff
@@ -105,6 +117,65 @@ impl Route {
...
+    /// We need to establish a priority. The lower of the returned number,
+    /// the more priority it has. This number is calculated based on the number of used
+    /// parameters, as fixed routes has more priority than parameted ones.
+    ///
+    /// To avoid collisions like `[id]/b.wasm` vs `/a/[id].js`. Every depth level will
+    /// add an extra +1 to the score. So, in case of `[id]/b.wasm` vs `/a/[id].js`,
+    /// the /a/b path will be managed by `[id]/b.wasm`
+    ///
+    /// In case it cannot manage it, it will return -1
+    pub fn can_manage_path(&self, url_path: &str) -> i32 {
```
ereslibre, line 140: The name of this function hints to return a boolean, I'd rename the function to something more descriptive.

src/router.rs
```diff
@@ -105,6 +117,65 @@ impl Route {
             })
             .collect()
     }
+
+    /// Check if the given path can be managed by this path. This was introduced
```
ereslibre, line 121:
```suggestion
    /// Check if the given path can be managed by this worker. This was introduced
```

src/router.rs
ereslibre: Also, instead of numbers with special semantic values (as in `-1` in this case), I think it's more clear to use an enum, like:

```rust
// Renamed can_manage_path to path_affinity as a possible alternative name
pub fn path_affinity(&self, url_path: &str) -> PathAffinity {
}

type PathAffinityDistance = i32;

enum PathAffinity {
    CannotManageEndpoint,
    CanManageEndpoint(PathAffinityDistance),
}
```

I might have got the example too far, but I hope it makes sense. In the end, the result is also performant, but I believe the source code is easier to read (no need to infer what is the meaning of special values as `-1`).

src/router.rs
```diff
@@ -144,6 +215,24 @@ fn is_in_public_folder(path: &Path) -> bool {
...
+        let route_score = item.can_manage_path(path);
+
+        if route_score != -1 && (best_score == -1 || best_score > route_score) {
+            best_score = route_score;
+            item
+        } else {
+            acc
+        }
+    })
+}
```
ereslibre, line 234: Isn't this going to produce always the same result? Unless I'm missing something, the `reduce` here is not "selecting" but summing all routes somehow into the accumulator, and returning that.

src/router.rs
```diff
@@ -179,6 +268,38 @@ pub fn format_prefix(source: &str) -> String {
 mod tests {
     use super::*;
 
+    #[test]
+    fn can_manage_path_scores() {
```
ereslibre, line 272: I think a set of tests checking the logic of `retrieve_best_route` would be nice.

ereslibre (commented):
Super cool feature and work! I have some comments regarding compatibility mainly with Windows, and other minor comments.

All in all, I think despite surprising to have brackets at the filesystem level, it is a useful feature when developing locally and prioritizing fast results.

src/router.rs
Angelmmiguel: Good point. I'm gonna refactor this based on your proposal 👍

docs/docs/features/dynamic-routes.md
```diff
@@ -0,0 +1,41 @@
+---
+sidebar_position: 3
+---
+
+# Dynamic routes
+
+Defining static routes may not be enough for some applications. You may need a worker to process URLs that includes identifiers that may change. **To create a worker associated with a dynamic route, include the route parameter in brackets when setting the worker filename**. For example, the worker `[id].js` will be able to reply to `/abc` or `/test` routes.
```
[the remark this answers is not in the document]
Angelmmiguel: Good point! I already updated it to make the documentation more clear 😄

src/router.rs
Angelmmiguel: Based on your other comment, I changed a bit this code. However, the result was properly calculated before as the `reduce` method returns the `acc` or `item` depending on the different conditions.

I fixed a different bug after implementing the tests as this method was always returning a route. `acc` is initialized to the first item, so even the first one cannot manage the route, it was selected. For this reason, I switched to the `fold` method so I could configure the initial value.

src/router.rs
Angelmmiguel: Totally agree! I just added the tests :)

ereslibre (approved):
🚢 it!

src/main.rs
```diff
@@ -52,118 +53,145 @@ struct Args {
...
+            .run(&req, body_str, store, vars)
+            .unwrap_or_else(|_| WasmOutput::failed());
+
+        let mut builder = HttpResponse::build(
+            StatusCode::from_u16(handler_result.status).unwrap_or(StatusCode::OK),
+        );
+        // Default content type
+        builder.insert_header(("Content-Type", "text/html"));
+
+        for (key, val) in handler_result.headers.iter() {
```
gzurl, line 172: Is iter() really needed? I think the preferred style is using a reference instead of explicit `.iter()`:
https://rust-lang.github.io/rust-clippy/master/index.html#explicit_iter_loop.

```suggestion
        for (key, val) in &handler_result.headers {
```

PR #56, opened 2023-01-10
Improve filesystem routing and path management

### LINKED ISSUE #3: Improve path management in wws

Angelmmiguel:
Refactor the logic to load the app routes from the filesystem. Before, we were using functions instead of related struct implementations. The code was not clear and `structs` were taking mixed responsibilities. For example, we had a public function to `retrieve_best_route` that takes `Routes` as a parameter. I converted this into a `Routes`'s method directly.

This is the list of changes I applied:

- Create a separate folder for the filesystem routing logic (`router`)
- Split `Routes` and `Route` structs into separate files
- Move public functions to `struct` implementations to avoid passing references around to provide a result
- Replace `glob` library with `wax`. This library provides a better support for glob patterns. For example, `glob` doesn't support multiple options like `*.{js,wasm}`
- Update the `main` entrypoint to use the new APIs. It didn't require many changes.

It closes #3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -390,6 +390,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "brownstone"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c5839ee4f953e811bfdcf223f509cb2c6a3e1447959b0bff459405575bc17f22"
+dependencies = [
+ "arrayvec",
+]
+
+[[package]]
+name = "bstr"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba3569f383e8f1598449f1a423e72e99569137b47740b1da11ef19af3d5c3223"
+dependencies = [
+ "lazy_static",
+ "memchr",
+ "regex-automata",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


ereslibre (approved):
LGTM, thanks @Angelmmiguel! Nonblocking: I do wonder if we want to revisit visibility of certain structs/functions.

Angelmmiguel:
> LGTM, thanks @Angelmmiguel! Nonblocking: I do wonder if we want to revisit visibility of certain structs/functions.

Great catch @ereslibre! I fixed the visibility of `normalize_path_to_url` and `is_in_public_folder` methods. Did you find any other that should be changed? 

Thanks!

ereslibre:
> Did you find any other that should be changed?

I was thinking in this part for this specific patch:

```diff
diff --git a/src/main.rs b/src/main.rs
index 7272f9f..403ae82 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -19,7 +19,7 @@ use actix_web::{
 };
 use clap::Parser;
 use data::kv::KV;
-use router::Routes;
+use router::routes::Routes;
 use runner::WasmOutput;
 use std::io::{Error, ErrorKind};
 use std::path::PathBuf;
diff --git a/src/router/mod.rs b/src/router/mod.rs
index 39679f8..9d0bed5 100644
--- a/src/router/mod.rs
+++ b/src/router/mod.rs
@@ -1,9 +1,6 @@
 // Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0

-pub mod files;
-pub mod route;
-pub mod routes;
-
-pub use route::Route;
-pub use routes::Routes;
+mod files;
+mod route;
+pub(super) mod routes;
```

However, we could double check all `pub` exports to identify which ones are really required to be public (I did run a `rg 'pub '`)

Angelmmiguel:
> > Did you find any other that should be changed?
> 
> I was thinking in this part for this specific patch:
> 

Good point. The only required entity is `Routes`. I will make it the only public entity in the `router` 👍

Angelmmiguel:
I rebased the changes with `main` to solve git conflicts due to `Cargo.lock`

Issue #66, opened 2023-01-20
Create a struct to manage runtime metadata files


Angelmmiguel:
The purpose of this struct is to act as a "Runtime Metadata Manager". The main tasks of this struct will be: 

- [x] Load metadata files for new runtimes from the local filesystem or a remote one
- [x] Download related resources and store them using the `runtime::Data` struct. We will need to add new methods to this one
- [x] Validate the different files using the associated checksums

This can be split in multiple structs / modules if required.

PR #72, opened 2023-01-23
feat: create the RuntimeMetadata and Repository structs. Refactor Runtimes

### LINKED ISSUE #64: Implement the runtime metadata file (toml)
These kind of files will be managed by `wws`, so it requires a representation in the source code. For consistency with the project, we chose the format as [`TOML`](https://toml.io/en/).

The purpose of them is to "teach" `wws` how to use certain runtime and which files can be run with it. For this reason, it must include information about the name, version and binary, but also polyfills, arguments and environment variables.

## Content

| Property | Type | Description |
| --- | --- | --- |
| spec_version | `u32` | The version of this file
| name | `String` | The runtime name |
| version | `String` | The runtime version number |
| binary | `String` | URL to download the binary
| binary_checksum | `String` | Binary checksum
| extensions | `String[]` | Associated file extensions to this runtime
| polyfill (optional) | `String` | An associated source file with polyfills. It will be mounted inside `lib` folder using WASI
| polyfill_checksum (optional) | `String` | The checksum to validate the polyfill file
| template (optional) | `String` | An optional template in case the source code needs to be modified. For example, to load a library or to add a function call. The provided code will be available in the {code} key
| template_checksum (optional) | `String` | The checksum to validate the template file
| args | `String[]` | Arguments to pass in the WASI context
| envs (optional) | `HashMap<String, String>` | Default environment variables
| folders (optional) | `HashMap<String, Object>` | A list of folders to mount with this runtime. It includes URL to download a compressed file, mount point and if decompression is required.

We may change these names or add some of them in the future.

It's related to #63

### LINKED ISSUE #65: Implement the runtime metadata file (toml)
These kind of files will be managed by `wws`, so it requires a representation in the source code. For consistency with the project, we chose the format as [`TOML`](https://toml.io/en/).

The purpose of them is to "teach" `wws` how to use certain runtime and which files can be run with it. For this reason, it must include information about the name, version and binary, but also polyfills, arguments and environment variables.

## Content

| Property | Type | Description |
| --- | --- | --- |
| name | `String` | The runtime name |
| version | `String` | The runtime version number |
| status | `Enum` | The status of the given runtime: `active`, `yanked`, `deprecated` |
| binary | `String` | URL to download the binary
| binary_checksum | `String` | Binary checksum
| extensions | `String[]` | Associated file extensions to this runtime
| polyfill (optional) | `String` | An associated source file with polyfills. It will be mounted inside `lib` folder using WASI
| polyfill_checksum (optional) | `String` | The checksum to validate the polyfill file
| template (optional) | `String` | An optional template in case the source code needs to be modified. For example, to load a library or to add a function call. The provided code will be available in the {code} key
| template_checksum (optional) | `String` | The checksum to validate the template file
| args | `String[]` | Arguments to pass in the WASI context
| envs (optional) | `HashMap<String, String>` | Default environment variables
| folders (optional) | `HashMap<String, Object>` | A list of folders to mount with this runtime. It includes URL to download a compressed file, mount point and if decompression is required.

We may change these names or add some of them in the future.

### LINKED ISSUE #71: Implement the index metadata file
The purpose of the `index.toml` metadata file is to provide simplify the installation of new runtimes in `wws`. This file will live in a repository and point to the different runtimes available on it. By default, `wws` will rely on the [WebAssembly Languages Runtime project](https://github.com/vmware-labs/webassembly-language-runtimes). However, any other file repository can be used instead in wws.

The file format will be `TOML` to be consistent with the other configuration files.

## Fields

| Name | Description |
| --- | --- |
| `version` | The version of the index file |
| `runtimes` | The list of runtimes. Every runtime contains all the fields defined at #65. |

Angelmmiguel:
Introduce the new `RuntimeMetadata` and `Repository` structs. They will deserialize the metadata from a `TOML` file, and then, it will be managed by the `RuntimeManager` (#66). In this PR, I also applied a final refactor by moving all the runtime code in a separate `runtimes` module.

For now, this code is quite abstract. I didn't implement any logic yet.

It closes #65, it closes #71


src/runtimes/metadata.rs
```diff
@@ -0,0 +1,59 @@
...
+    extensions: Vec<&'a str>,
+    /// Arguments to pass to the Wasm module via WASI
+    args: Vec<&'a str>,
+    /// A list of environment variables that must be configured
+    /// for the runtime to work.
+    envs: Option<HashMap<&'a str, &'a str>>,
+    /// The URL to pull the module binary
+    binary: &'a str,
+    /// The checksum to validate the given binary (SHA256)
+    binary_checksum: &'a str,
```
ereslibre, line 37: This could be an enum that only holds the SHA256 option as of today, like this:

```rust
binary: &'a RuntimeBinary,
```

and then,

```rust
pub struct RuntimeBinary<'a> {
  url: &'a str,
  checksum: Checksum<'a>,
}
```

```rust
pub enum Checksum<'a> {
  Sha256(&'a str),
}
```

So that we could support different checksums in the future if desired in an easier way.

src/runtimes/metadata.rs
```diff
@@ -0,0 +1,59 @@
...
+
+// TODO: Remove it when implementing the manager
+#[allow(dead_code)]
+impl<'a> RuntimeMetadata<'a> {
+    /// Reads and parses the metadata from a slice of bytes. It will return
+    /// a result as the deserialization may fail.
+    pub fn from_slice(data: &'a [u8]) -> Result<Self> {
+        toml::from_slice::<RuntimeMetadata>(data)
+            .map_err(|_| anyhow!("wws could not deserialize the runtime metadata"))
+    }
```
ereslibre, line 58: Do we expect third parties to integrate wws as an API at some point? If so, I think we should start returning `Result` with our own custom Error types, so that callers can have this information in a more streamlined way.

ereslibre (approved):
LGTM; thanks @Angelmmiguel! Non blocking improvement suggestions.

src/runtimes/metadata.rs
Angelmmiguel: Good point. This will ensure we have a single entity per file and it's supported by `TOML`. Will apply these changes 👍

Angelmmiguel: I've just applied your suggestions. I also added some tests to ensure deserializing is working

src/runtimes/metadata.rs
Angelmmiguel: I created this issue for tracking this effort: https://github.com/vmware-labs/wasm-workers-server/issues/73

Issue #67, opened 2023-01-20
Implement the `.wws.toml` file


Angelmmiguel:
The `.wws.toml` file contains information about the current project and the required language runtimes. For now, this file will be auto-generated and we don't expect developers to modify it. 

It includes the pointer to the metadata file and it's checksum. When this file is committed to the repo, `wws` will detect it, download the runtime metadata and install the missing ones. 

We may add this as part of the default `wws` command or add a separate one just for installing missing runtimes.

## Content / Format

For the `v1`, the file will contain only the basics about the installed runtimes / repositories. I wouldn't go further with the implementation as at this point, `wws` doesn't really need more data.

So, the format for the `.wws.toml` file will be:

```toml
# Refers to the version of this file
version = 1

[[repository]]
name = "wlr"

[[repository.runtimes]]
# Runtime metadata
name = "ruby"
version = "3.2.0"
binary = { url = "...", checksum = "...", filename = "..." }
# ...
[[repository.runtimes]]
# Runtime metadata
name = "python"
version = "3.2.0"
binary = { url = "...", checksum = "...", filename = "..." }
# ...
```

At this point, you may noticed similarities and differences with other popular project definition files like `package.json` or `Cargo.toml`. In both Rust and NodeJS, the project definition differs from their lock files (`Cargo.lock` and `package-lock.json`). The main reason is to provide an friendly file for developers (definitions) and a full spec with all the requirements for the tools (lock files).

However, in this case we don't expect you to install many language runtimes. Also, these runtimes are self-contained. A single language runtime will be defined by an unique block. They don't have dependencies you need to fetch and refer in a lock file.

For this reason, the `.wws.toml` will include both definitions (runtime names and version) and "lock" metadata (URLs, checksums, etc). If in the future this file grows and it's not maneagable anymore, we will open a discussion about splitting it.

PR #74, opened 2023-01-25
feat: create methods to manage runtimes

### LINKED ISSUE #66: Create a struct to manage runtime metadata files

Angelmmiguel:
This PR includes several refactors and new features to manage language runtimes. In general, this PR prepares the code for the complete feature. The main changes applied to this PR are:

* Create a new set of utilities for fetching data. It uses [`reqwest`](https://docs.rs/reqwest/0.11.14/reqwest/index.html#) in async mode. For the validation side, it receives a `Checksum` instance to validate against the content.
* Refactor the `Data` struct that was used for language runtimes to install transient files in a `.wws` directory. Now, we also need to download and store runtime files such as wasm modules, polyfills and templates. For this reason, I moved the `.wws` folder management into a global `Store` struct. I also fixed the routes to work on Windows too.
* Create a "manager" set of utilities to initialize runtimes, install and download associated files. It uses the aforementioned `Store` struct to create the files in the correct location.
* Merge all repository / runtime metadata structs into a common `runtimes/metadata.rs` file. This way is less confusing to understand the scope of those structs.
* Unfortunately, I run into `lifetimes` issues with Serde and the `Repository -> Runtime -> RemoteFile` struct. In the end, I opted for making all owned types and revisit it in the future 😞
* Add a `filename` field to the `RemoteFile` struct. 

## Usage experience

```
println!("🗺  Fetching index!");
let repo = Repository::from_remote_file(
    "https://raw.githubusercontent.com/Angelmmiguel/wws-index-test/main/index.toml",
).await.unwrap();

let runtime = repo.runtimes.first().unwrap()

println!("🗺  Installing  runtime...");
println!("{}", runtime);
install_runtime("wlr", runtime).await.unwrap();
```

And then, you can inspect the `.wws` folder:

```
~/W/o/wws ❯❯❯ tree .wws
.wws
├── runtimes
│   ├── rust
│   └── wlr
│       └── ruby
│           └── 3.2.0+20230118-8aec06d
│               ├── poly.rb
│               ├── ruby.wasm
│               └── template.txt
└── workers
    └── js
        └── 961b7b706f36e745dc7c3306a2a41214ddac3ad41a38fa8da6a2460058235420
            └── index.js
```

It closes #66


Angelmmiguel:
Added a new commit to prepare the metadata to be serialized. This required to update `TOML` to v0.6.0 to avoid an issue regarding serialization of values and tables (see https://github.com/toml-rs/toml-rs/issues/142). 

On this version, `toml` removed the support for `from_slice`, so I had to use the `from_str` method instead (see [v0.6.0 changelog](https://github.com/toml-rs/toml/blob/main/crates/toml/CHANGELOG.md#060---2023-01-23)).

The reason why I'm adding it here is to prepare the metadata to be serialized on #67

src/config.rs
```diff
@@ -46,9 +46,10 @@ impl Config {
     /// namespace = "todos"
     /// ```
     pub fn try_from_file(path: PathBuf) -> Result<Self, String> {
-        let contents = fs::read(&path).expect("The configuration file was not properly loaded");
+        let contents =
+            fs::read_to_string(&path).expect("The configuration file was not properly loaded");
```
ereslibre, line 50: Don't we want to bubble the result? I did not realize on a previous pass, we are panicking with `expect` here.

src/runtimes/metadata.rs
```diff
@@ -16,53 +56,117 @@ use std::collections::HashMap;
...
-#[allow(dead_code)]
-impl<'a> RuntimeMetadata<'a> {
-    /// Reads and parses the metadata from a slice of bytes. It will return
-    /// a result as the deserialization may fail.
-    pub fn from_slice(data: &'a [u8]) -> Result<Self> {
-        toml::from_slice::<RuntimeMetadata>(data)
-            .map_err(|_| anyhow!("wws could not deserialize the runtime metadata"))
+// Implement the Display trait to runtime metadata properly in the CLI
+impl fmt::Display for Runtime {
+    /// Prints the runtime as a row for a list of runtimes
```
ereslibre, line 86: I think is file, but maybe we can derive Display for Runtime for general println!() statements, wherever we might need them, and for reporting a list of runtimes and similar stuff we could use `ascii_table` or `prettytable_rs`, so that they are nicely aligned and so on.

Angelmmiguel:
Added a new commit to fix the location of the `.wws` folder. It was using the current path instead of the project path. For example, if we run the following command:

```
wws ./examples/js-basic
```

Before this commit, the `.wws` would be located in `./.wws`. However, the correct path for this folder is the project one: `./examples/js-basic/.wws`. The last commit fixes that behavior.

ereslibre (approved):
LGTM, minor comments

src/config.rs
Angelmmiguel: You're totally right. I'm gonna update it.

src/runtimes/metadata.rs
Angelmmiguel: Good point. I'm gonna remove this for now and will work on the final implementation when showing the list of runtimes in the command.

Issue #63, opened 2023-01-20
Adding multi-language support (multiple language runtimes)


Angelmmiguel:
_This issue tracks an ongoing effort. The implementation will be done in the related issues you can find at the end_

Wasm Workers Servers aim to create serverless applications based on different languages. As a starting point, we added support for Wasm modules (compiled from Rust) and JavaScript (by embedding the [QuickJS](https://bellard.org/quickjs/) runtime). All these workers run inside the WebAssembly sandbox.

For v1.0.0, the goal is to extend the number of languages by allowing different language runtimes. Even though the size of other runtimes like Ruby and Python are not big (see our [python](https://github.com/vmware-labs/webassembly-language-runtimes/releases/tag/python%2F3.11.1%2B20230118-f23f3f3) and [ruby](https://github.com/vmware-labs/webassembly-language-runtimes/releases/tag/ruby%2F3.2.0%2B20230118-8aec06d) releases), including them in the main `wws` is not an option.

In addition to that, we want to make `wws` extensible. Any developer may implement new language runtimes and use them. For this reason, `wws` will allow to import runtimes from different sources.

## CLI experience

In this section, I will explore the desired CLI experience for this feature. All the examples assume a project has two or more workers. One is a "native" worker (.wasm) file that doesn't require any runtime. The other file requires a runtime to run (`.rb` or `.py`). 

```
$ tree ./examples/multi-lang 
./examples/ multi-lang 
├── hello.rb 
├── basic.toml 
└──  basic.wasm 
```

### Repositories

A repository is an index file that points to different runtimes. It can be hosted in a git repository or in any service that can expose files. `wws` will retrieve the index file from the given repository. This index file contains pointers to the available runtimes. The repository is configurable via arguments, so anyone can develop their own repository. If no `--repo` parameter is provided, the [WebAssembly Language Runtimes](vmware-labs/webassembly-language-runtimes) will be the default repo. 

The CLI experience would be: 

```
$ wws runtimes list 
NAME            VERSION         EXTENSION       BINARY          STATUS 
ruby            3.2.0           .rb             https://...     not installed 
ruby-std        3.2.0           .rb             https://...     not installed 
python          2.3             .py             https://...     not installed 

$ wws runtimes install ruby-std 3.2.0 
Pulling the runtime binary... 
Installing it in ./wws/runtimes/wlr/ruby-std/3.2.0/... 
Validating checksum... 
Done!

$ wws ./examples/multi-lang 
```

[...]

PR #75, opened 2023-01-26
feat: create and manage the .wws.toml file and its metadata

### LINKED ISSUE #67: Implement the `.wws.toml` file

Angelmmiguel:
The `.wws.toml` file includes information about the required runtimes for a specific project. It will allow other developers to quickly install these dependencies when pulling a new `wws` project. 

In this PR, I moved the existing `Config` struct that referenced to workers configuration into the `workers` module. Then, I created a new `Config` fiel to refer to the `.wws.toml` file. The struct implements some of the methods thart will be required to interact with this file like:

- `load`
- `save`
- `save_runtime`
- `remove_runtime`

## Example

```rust
let mut config = config::Config::load(&args.path).unwrap();
let repo = Repository::from_remote_file(
    "https://raw.githubusercontent.com/Angelmmiguel/wws-index-test/main/index.toml",
)
.await
.unwrap();

let runtime = repo.runtimes.first().unwrap();

config.remove_runtime("wlr", runtime);
config.save(&args.path).unwrap();
```

It closes #67

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ target
 examples/*.toml
 .DS_Store
 .wws
+.wws.toml
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,109 +1,108 @@
 // Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::data::kv::KVConfigData;
+use crate::runtimes::metadata::Runtime;
 use anyhow::{anyhow, Result};
-use serde::{Deserialize, Deserializer};
-use std::collections::HashMap;
-use std::path::PathBuf;
-use std::{env, fs};
-use toml::from_str;
+use serde::{Deserialize, Serialize};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 
-/// Workers configuration. These files are optional when no configuration change is required.
-#[derive(Deserialize, Clone)]
+/// Config file name
+const CONFIG_FILENAME: &str = ".wws.toml";
+
+/// Loads the data from the Project definition file or .wws.toml.
+/// This file contains information about the different runtimes
+/// required for this project. You can think of those as dependencies.
+///
+/// If your project requires to run workers using any interpreted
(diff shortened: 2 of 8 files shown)


Angelmmiguel:
This API is not exposed yet to devs, so I'm gonna merge these changes to continue with #63. Feel free to add any comment to this PR and I will apply them in a new one 😄

PR #77, opened 2023-01-26
fix: add missing OpenSSL dependency for building wws 

Angelmmiguel:
I tried to fix this error by installing the required dependencies following the recommendation in the [openssl-sys crate](https://docs.rs/openssl/0.10.45/openssl/index.html#automatic). However, this only works for the `gnu` tooling. Since we're using `musl` for compiling static binaries, we cannot rely on this solution.

I tested the usage of rusttls instead of openssl. I couldn't make it work in Windows ARM64, as one of the dependencies ([ring](https://github.com/briansmith/ring)) is not compiling on that environment with the latest version. The team did an amazing job of supporting this platform and it's already fixed in upstream ([See this PR](https://github.com/briansmith/ring/pull/1339)). However, this version is not published yet. It seems [they're planning to release the 0.16.21 it soon](https://github.com/briansmith/ring/issues/1551). Until this version is published, unfortunately we won't be able to swich to `rusttls`.

In the end, I used the [`vendored` feature from the `openssl` crate](https://docs.rs/openssl/latest/openssl/#vendored) in both `musl` target. This feature compiles the OpenSSL library from source. I tested and it's properly working. 

You can find here a successful run, as I removed the PR trigger in the latest commit: https://github.com/vmware-labs/wasm-workers-server/actions/runs/4016745514

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -17,6 +17,7 @@ jobs:
       run: cargo test
   build:
     strategy:
+      fail-fast: false
       matrix:
         build: [linux, windows, macos]
         arch: [x86_64, aarch64]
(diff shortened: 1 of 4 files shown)


assambar (approved):
Great research and findings! Thanks for the detailed explanation!

ereslibre (approved):
Thanks for the great research @Angelmmiguel! LGTM

PR #88, opened 2023-02-13
feat: add the initial repository index

### LINKED ISSUE #85: Create the initial repository index
From v1.0.0, `wws` relies on external repositories to fetch the available language runtimes. By default, this project will host the initial metadata:

- Index
- Polyfills
- Wrappers

We will host it as a static assets, while relying on GitHub releases for the different artifacts. Initially, we will generate the index manually based on the [WebAssembly Language Runtimes](https://github.com/vmware-labs/webassembly-language-runtimes/) project. We will automate this process in the future by adding several automations in both Language Runtimes and Wasm Workers Server projects.

For the URLs, we plan to segment it by version even it's included in the index spec. This will allow in the future to add redirections and any required change to maintain a better backward compatibility in case we need to change it. So, the final URL will be:

* https://workers.wasmlabs.dev/repository/v1/index.toml

`wws` will use this metadata unless the user specify a different repository. The name of this repo will be `wasmlabs`.

## Versioning

In the repository metadata we have to version both runtimes and also other files such as polyfills and wrappers. These files have two different, but related versions. In general, a polyfill will work for the major version of a language runtime. However, we may add new features that requires to update these polyfills.

Since the runtime metadata is stored locally in the `.wws.toml` file, updating a related file (polyfill or wrapper) and the `index.toml` entries will break developer environments that have references to the previous file version. For this reason, we musn't change polyfills and wrappers after releasing them. 

If we need to do so, we will: 

- Create the new files under a different revision (for example, `/repository/v1/files/ruby/3-1/poly.rb`)
- Create a new release pointing to the new files
- Mark the previous one as yanked

This will ensure new installation gets the latest version and existing ones work properly. We will add a warning message asking the users to update to the latest version if this happens.

Angelmmiguel:
In this PR, I'm introducing the initial repository index based on the current releases from the [WebAssembly Language Runtimes](https://github.com/vmware-labs/webassembly-language-runtimes) project. The polyfills and wrappers are stored as static files in this repository, so we're only referencing the runtime binaries.

All these files will be bundled as part of the documentation site, so the URLs are easy to access and remember.

There are still some pending items before we can merge this PR:

- [x] Release the Python and Ruby builds with bundled standard libraries
- [x] Update the entries in the `index.toml` file

It closes #85


assambar (approved):

ereslibre (approved):
Thanks @Angelmmiguel! LGTM 👏

Angelmmiguel:
Yeah! We all put a lot of work here 😄

PR #90, opened 2023-02-13
feat: check missing runtimes before loading workers

### LINKED ISSUE #69: Check required lang runtimes when running `wws` (from `.wws.toml`)
As part of the `wws` initialization, it will retrieve the data from the `.wws.toml` file (#67). Then, it will use the runtime manager (#66) to check if the required runtimes are installed. If they are missing, it will install them by default in the current folder.

This step is required as runtimes may add more supported extensions to `wws`. For example, by default `wws` cannot support workers written in Python. However, they may add a Python runtime and `wws` will start managing `.py` files as workers.

If this step fails for any reason, `wws` will show a warning and will continue with the remaining supported extensions.

Angelmmiguel:
When a user downloads an existing project, the required runtimes won't be installed. As part of the process, they need to download it using `wws runtimes install` as other similar tools like `NodeJS`. We want to provide a hint about this behavior so `wws` doesn't fail and it shows useful information.

From now on, `wws` will look for missing runtimes when running the `run` command (default one). If there're missing runtimes, it will show a message and will ask users to install it using `wws runtimes install`. In the future, we will evaluate if installing missing runtimes should be part of the regular process.

In addition to that, now `wws` will ignore the extensions of not installed runtimes.

```
./wws .
⚠️  Required language runtimes are not installed. Some files may not be considered workers
⚠️  You can install the missing runtimes with: wws runtimes install
⚙️  Loading routes from: .
🗺  Detected routes:
```

It closes #69

### DIFF
--- a/src/runtimes/manager.rs
+++ b/src/runtimes/manager.rs
@@ -132,18 +132,15 @@ pub fn check_runtime(project_root: &Path, repository: &str, runtime: &RuntimeMet
 
 /// Check if there're missing runtimes based on the current configuration
 pub fn check_runtimes(project_root: &Path, config: &Config) -> bool {
-    let mut missing = false;
-
-    'outer: for repo in &config.repositories {
+    for repo in &config.repositories {
         for runtime in &repo.runtimes {
             if !check_runtime(project_root, &repo.name, runtime) {
-                missing = true;
-                break 'outer;
+                return true;
             }
         }
     }
 
-    missing
+    false
 }
 
 // Install a given runtime based on its metadata
--- a/src/runtimes/manager.rs
+++ b/src/runtimes/manager.rs
@@ -130,6 +130,22 @@ pub fn check_runtime(project_root: &Path, repository: &str, runtime: &RuntimeMet
     binary && template && polyfill && wrapper
 }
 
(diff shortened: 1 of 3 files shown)


src/runtimes/manager.rs
```diff
@@ -130,6 +130,22 @@ pub fn check_runtime(project_root: &Path, repository: &str, runtime: &RuntimeMet
...
 }
 
+/// Check if there're missing runtimes based on the current configuration
+pub fn check_runtimes(project_root: &Path, config: &Config) -> bool {
+    let mut missing = false;
+
+    'outer: for repo in &config.repositories {
+        for runtime in &repo.runtimes {
+            if !check_runtime(project_root, &repo.name, runtime) {
+                missing = true;
```
assambar, line 140: Not a Rust guy, but this got me wondering why not return true here and reuturn false below, without modifying an additional variable?

assambar (approved):
That's a good improvement!

src/runtimes/manager.rs
Angelmmiguel: You're totally right. In this case, we can return the `true` value early 😄

PR #97, opened 2023-02-20
fix: avoid exiting as Tokio needs to manage open threads. Fix typo

### LINKED ISSUE #96: Manage errors properly when checksums don't match
### Describe the bug

When a checksum doesn't match after downloading a file, the CLI shows an error. However, it doesn't stop the execution and it fails while trying to save the file.

The error message is:

```
⚙️  Fetching data from the repository...
🚀 Installing the runtime...
❌ There was an error installing the runtime from the repository
👉 The checksums dont not match
thread '<unnamed>' panicked at 'cannot access a Thread Local Storage value during or after destruction: AccessError', /rustc/fc594f15669680fa70d255faec3ca3fb507c3405/library/std/src/thread/local.rs:422:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
```

We should avoid panicking and stopping the process at this point.

### Reproduction steps

1. Create an index with a wrong checksum
2. Try to install the given runtime

### Expected behavior

The error message doesn't contain a typo and it doesn't panik.

### Additional context

_No response_

Angelmmiguel:
Properly manage errors when running commands by using the `Result` return value. Before, I was exiting the CLI earlier and that was causing the app to panik. The reason is that, Actix uses Tokio under the hood for managing async code. Tokio need to manage the entire application to ensure the threads are properly finished. By exiting early, we force the program to panik.

I also fixed a small typo.

## Result

```
./wws install ruby latest
⚙️  Fetching data from the repository...
🚀 Installing the runtime...
❌ There was an error installing the runtime from the repository
👉 The checksums don't match
Error: Custom { kind: InvalidData, error: "" }
```

I couldn't remove the "Error: Custom" message for now. However, this is much better than the previous output so I'll investigate why this is happening in a future version.

It closes #96

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,7 +30,6 @@ use data::kv::KV;
 use router::Routes;
 use std::io::{Error, ErrorKind};
 use std::path::PathBuf;
-use std::process::exit;
 use std::{collections::HashMap, sync::RwLock};
 use workers::wasm_io::WasmOutput;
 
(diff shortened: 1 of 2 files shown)


Angelmmiguel:
Since I plan to revisit how to improve this in the future, I will merge these changes.

Issue #73, opened 2023-01-24
Improve overall error capturing


Angelmmiguel:
_This issure requires a better definition once we plan to work on it_

`wws` heavily relies on the [`anyhow`](https://crates.io/crates/anyhow) for a simple error management. In certain cases, we're capturing the errors and mapping it to more appropiate messages. However, this focuses on the errors we want to show in the CLI.

We want to create a set of specific error types to capture and manage them in code properly. Instead of bubbling generic errors, we plan to throw a set of specific ones we can manage later.

PR #106, opened 2023-03-09
feat: split wws code in multiple crates

### LINKED ISSUE #102: Split the project codebase in multiple crates to simplify development and consumption

Currently, the `wws` codebase is defined as a single crate (under `src`). This approach worked when compiling `wws` as a CLI, since the output is a single artifact. However, for consuming `wws` as a library, this approach forces to compile almost the entire project just to consume a few resources.

For this reason, we want to split the project into several small crates that can be consumed independently. All of them will belong to the same workspace. It simplifies the way we can consume `wws` as a library as well as ensuring the code is more organized. 

After this change, we will have a `crates` folder with the individual pieces of this project: `store`, `runtimes`, `workers`, etc.

Exposing the entire project as a library (adding the proper configuration to Cargo.toml) works, but it requires to expose all the pieces and to compile the entire project when consuming it.

### Additional context

_No response_

Angelmmiguel:
The purpose of this PR is to split the current project into separate crates. Currently, all the code is split in modules inside the `./src` folder. This is causing some issues:

* When importing `wws` as a library, it needs to compile the entire project. This includes `clap` and other libraries that are only related to the `wws` CLI
* Some parts of the code have cyclic dependencies (X relies on Y and the opposite). This is difficult to spot when all the code is in the same crate. By splitting it into multiple crates, we define a clear responsibility and API for the different crates
* Applying changes and fixes to specific crates is simpler to do it on the entire project
* We can use the Cargo `workspace` tooling to manage and share dependencies easily

Now it's a great moment to perfrom these changes as we just released a new version and we're planning to start using `wws` as a library.

For now, I identified these crates:

- [x] `config`: manage the `wws` global configuration (`.wws.toml` file)
- [x] `data-kv`: manage the KV in-memory store we ship with wws
- [x] `router`: define the routes by loading files from the local filesystem. It also initializes the workers
- [x] `runtimes`: initalize and configure `wasmtime` to run workers' source code. 
- [x] `runtimes-manager`: interacts with the remote repository to install and uninstall language runtimes (ruby, python, etc)
- [x] `server`: run the final HTTP server users can query
- [x] `store`: manage the temporary store for the runtime and worker files / folders
- [x] `worker`: initialize all the required resources to run a worker from the given file. It also calls the `runtime` to return a response

It closes #102


crates/config/src/lib.rs
```diff
@@ -91,6 +93,19 @@ impl Config {
...
+    pub fn check_runtimes(&self, project_root: &Path) -> bool {
+        for repo in &self.repositories {
+            for runtime in &repo.runtimes {
+                if !check_runtime(project_root, &repo.name, runtime) {
+                    return true;
+                }
+            }
+        }
+
+        false
```
ereslibre, line 106: This can be replaced with `Iter::any`, I think this way it's more idiomatic.

However, I would rename this function, because I would expect a check returning a bool to be fine, and it's the other way around. Ideally instead of a bool, an enum does not let space for guessing. If we want to keep the bool, I would rename this function to something like `is_missing_any_runtime` or something that conveys this information.

crates/data-kv/src/lib.rs
```diff
@@ -55,31 +51,11 @@ impl KV {
...
-    pub fn new(namespace: String) -> Self {
+impl Default for KV {
+    /// Creates a new KV instance. It initializes the K/V stores to an empty HashMap
+    fn default() -> Self {
         Self {
-            namespace,
-            store: HashMap::new(),
+            stores: HashMap::new(),
         }
     }
```
ereslibre, line 60: We can just `#[derive(Default)]` on `KV`, given HashMap also implements this trait (https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-Default-for-HashMap%3CK%2C%20V%2C%20S%3E)

crates/runtimes-manager/src/lib.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
ereslibre, line 2: Year out of date, I would remove these headers that are on some files and not on others, given we already have a project-wide license.

crates/runtimes-manager/src/metadata.rs
```diff
@@ -68,6 +60,19 @@ impl Repository {
     }
 }
 
+impl FromStr for Repository {
+    type Err = anyhow::Error;
```
ereslibre, line 64: If this is meant to be a consumable library by third party crates (and ours as consumers too), I would avoid anyhow. We can improve this later though.

The idea for crates that are meant to be consumed as libraries is to define our own error types, and let third parties use those error types, whereas anyhow is great when the code reaches the code that is closer to the user (e.g. at CLI or near-CLI levels).

crates/server/src/handlers/worker.rs
```diff
@@ -0,0 +1,118 @@
...
+        // than dynamic routes. However, I cannot set the static assets as the first service
+        // as it's captures everything.
+        if route.is_dynamic() {
+            if let Ok(existing_file) = handle_assets(&req).await {
+                return existing_file.into_response(&req);
+            }
+        }
+
+        // Let's continue
+        let body_str = String::from_utf8(body.to_vec()).unwrap_or_else(|_| String::from(""));
```
ereslibre, line 55: Do we want to treat the body as a valid utf-8 encoded string? I think it would be better to treat it as a `Vec<u8>`.

crates/server/src/handlers/worker.rs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            Ok(output) => (output, true),
+            Err(_) => (WasmOutput::failed(), false),
+        };
+
+        let mut builder = HttpResponse::build(
+            StatusCode::from_u16(handler_result.status).unwrap_or(StatusCode::OK),
+        );
+        // Default content type
+        builder.insert_header(("Content-Type", "text/html"));
```
ereslibre, line 90: My previous comment can be ignored if we force `text/html` content-type though. Just a question: do we intend to do some kind of mime type detection and forward that, or let this to the worker completely?

crates/server/src/handlers/worker.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        for (key, val) in handler_result.headers.iter() {
+            // Note that QuickJS is replacing the "-" character
+            // with "_" on property keys. Here, we rollback it
+            builder.insert_header((key.replace('_', "-").as_str(), val.as_str()));
+        }
```
ereslibre, line 96: This could be part of the worker logic, kind of an operation to 'amend the http response' at the end before we return it. I say so, because this might be undoing some valid headers meant to contain `_`.

crates/server/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+
+/// Initializes an actix-web server based on the given configuration and
+/// path. It will configure the different handlers to manage static
+/// assets and workers.
+pub async fn serve(
+    root_path: &Path,
+    base_routes: Routes,
+    hostname: &str,
+    port: u16,
+) -> Result<Server> {
```
ereslibre, line 34: In general with `Result` and our exported library-functions and methods, I would define our own error type and use that, so that the consumer has richer types when things error out. Anyhow for this kind of usage flattens the errors into one and is harder to consume as a library.

crates/server/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+    // Initializes the data connectors. For now, just KV
+    let data = Data::new(RwLock::new(DataConnectors { kv: KV::default() }));
```
ereslibre, line 36:
```suggestion
    let data = Data::new(RwLock::new(DataConnectors::default()));
```

ereslibre (commented):
Thanks @Angelmmiguel! It's awesome to see how this code evolves.

crates/runtimes-manager/src/lib.rs
Angelmmiguel: Yeah, it's easy to keep these comments outdated :/. As far as I know, we should keep them. Let's discuss this in a separate thread.

crates/runtimes-manager/src/metadata.rs
Angelmmiguel: That's a good point. I don't have much experience with `anyhow` and Rust error handling, so I would research about it. I'll create a separate task to work on that 👍

crates/server/src/lib.rs
Angelmmiguel: Totally agree with this. I will create a separate task to revisit the error management in the project 👍

crates/server/src/handlers/worker.rs
Angelmmiguel: This is the default `content-type` for the worker output. And yes, it's the worker the one in charge of setting the right content-type if required. For example, I recently set a worker to return a `protobuf` content type.

Note that for [static assets](https://workers.wasmlabs.dev/docs/features/static-assets), `wws` identifies the content-type. But that's a different feature that doesn't involve the Wasm modules / workers.

crates/server/src/handlers/worker.rs
Angelmmiguel: Hmm that's a good point. The `wws-worker` crate is already tight to `actix-web` as it processes the inputs. I tend to think the `wws-worker` create shouldn't care of an `HttpRequest` / `HttpResponse` and just work with `WasmInput` / `WasmOutput`. The opposite solution is your suggestion. I agree we should either go in one direction or the other, but this is a mix.

I would think on this. To simplify the refactor, I may create a separate task to revisit this specific case 👍

crates/config/src/lib.rs
Angelmmiguel: Done! Thanks for the suggestion :)

crates/runtimes-manager/src/metadata.rs
Angelmmiguel: Related issue #73

crates/server/src/lib.rs
Angelmmiguel: Related issue #73

ereslibre (approved):
👏 🚢 it!

crates/server/src/handlers/worker.rs
Angelmmiguel: Related to #110

PR #126, opened 2023-05-05
Assambar/47 ignore option

### LINKED ISSUE #47: Add an option to ignore certain folders
When running `wws`, it will fetch all the files in the given directory and subfolders. Sometimes, we may run the CLI in a folder, but we don't want to check all the folders inside it. For example, we want `wws` to be able to ignore `node_modules` folders in certain cases.

To make it configurable, we plan to add a new option to the CLI called `--ignore`. This `--ignore` option can be used multiple times to ignore several folders:

```
wws --ignore node_modules --ignore tmp .
```

Clap supports already supports this feature. The complexity of this issue comes from:

* The support of Windows and Unix systems
* Usage of `Glob` crate

Note that if we use the same approach we do for skipping the `public` folder, `Glob` will still list all the files in those folders. Folders lik `node_modules` can cause a performance issue as they may contain a big number of supported files (`.js`).

### LINKED ISSUE #103: Display a message with the number of detected workers before processing them

By default, `wws` looks for workers in the given path and any directory inside. In certain projects, you may have some extra folders like python `venv` or NodeJS modules. You don't want these folders to be processed and this will be covered by #47.

However, this issue is not easy to notice. As discussed in #101, the CLI just hangs while trying to load all the files in the given directory.

To inform the user about this, we can add a new message before start processing the workers:

```
X files are identified as workers. Start processing them
```

This message will help to spot these issues as users will notice it when waiting for `wws` to process the workers.

_No response_

### Additional context

_No response_

assambar:
- added `--ignore` option that accepts full glob patterns (one can exclude files as well as folders)
 - refactored `router/files` to remove dependency from Config, also added a few more unit tests
 
Here is an example run from the repo's main folder
```
alexandrov:~/work/wasm-workers-server 05.05 15:05:47\>target/debug/wws . --ignore "src/**/*;target/**/*;examples/**/*;docs/**/*;crates/**/*;kits/**/*"
⚙️  Loading routes from: .
🗺  Detected 4 routes:
    - http://127.0.0.1:8080/tests/data/params/sub/[id]
      => ./tests/data/params/sub/[id].wasm
    - http://127.0.0.1:8080/tests/data/params/fixed
      => ./tests/data/params/fixed.wasm
    - http://127.0.0.1:8080/tests/data/params/[id]
      => ./tests/data/params/[id].wasm
    - http://127.0.0.1:8080/tests/data/params/[id]/fixed
      => ./tests/data/params/[id]/fixed.wasm

alexandrov:~/work/wasm-workers-server 05.05 15:05:55target/debug/wws . --ignore "{src,target,examples,docs,crates,kits}/**/*;**/*fixed*"
⚙️  Loading routes from: .
🗺  Detected 2 routes:
    - http://127.0.0.1:8080/tests/data/params/sub/[id]
      => ./tests/data/params/sub/[id].wasm
    - http://127.0.0.1:8080/tests/data/params/[id]
      => ./tests/data/params/[id].wasm

```

It closes #47. It closes #103

### DIFF
--- a/crates/router/src/files.rs
+++ b/crates/router/src/files.rs
@@ -81,6 +81,7 @@ mod tests {
     use super::*;
 
     use std::collections::HashSet;
+    use path_slash::PathBufExt as _;
 
     #[test]
     fn walk_default_ignore() {
@@ -88,15 +89,15 @@ mod tests {
         let files = Files::new(Path::new("tests/data/files"), &config);
 
         let mut expected = HashSet::new();
-        expected.insert("tests/data/files/examples.js".to_string());
-        expected.insert("tests/data/files/index.js".to_string());
-        expected.insert("tests/data/files/public.js".to_string());
-        expected.insert("tests/data/files/examples/public.js".to_string());
-        expected.insert("tests/data/files/examples/index/index.js".to_string());
+        expected.insert(PathBuf::from_slash("tests/data/files/examples.js"));
+        expected.insert(PathBuf::from_slash("tests/data/files/index.js"));
+        expected.insert(PathBuf::from_slash("tests/data/files/public.js"));
+        expected.insert(PathBuf::from_slash("tests/data/files/examples/public.js"));
+        expected.insert(PathBuf::from_slash("tests/data/files/examples/index/index.js"));
 
         let mut actual = HashSet::new();
         for entry in files.walk() {
-            actual.insert(String::from(entry.path().to_string_lossy()));
+            actual.insert(PathBuf::from_slash(String::from(entry.path().to_string_lossy())));
         }
 
         assert_eq!(expected, actual);
--- a/crates/router/src/files.rs
+++ b/crates/router/src/files.rs
@@ -4,8 +4,6 @@
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use wax::{Glob, WalkEntry};
-use wws_config::Config;
-use wws_runtimes_manager::check_runtime;
 use wws_store::STORE_FOLDER;
 
 const IGNORE_PATH_PREFIX: &str = "_";
@@ -24,33 +22,15 @@ pub struct Files {
 }
 
 impl Files {
+    const DEFAULT_EXTENSIONS: [&str; 2] = ["js", "wasm"];
+
     /// Initializes a new files instance. It will detect
     /// relevant resources for WWS like the public folder.
-    pub fn new(root: &Path, config: &Config) -> Self {
-        let mut extensions = vec![String::from("js"), String::from("wasm")];
-
-        for repo in &config.repositories {
-            for runtime in &repo.runtimes {
-                for ext in &runtime.extensions {
-                    if check_runtime(root, &repo.name, runtime) && !extensions.contains(ext) {
-                        extensions.push(ext.clone());
-                    }
-                }
-            }
-        }
-
-        let include_pattern: String = format!("**/*.{{{}}}", extensions.join(","));
-
-        let default_ignore_patterns = vec![
-            "**/public/**".to_string(),
-            format!("**/{}/**", STORE_FOLDER),
-            format!("**/{}*/**", IGNORE_PATH_PREFIX),
-        ];
-
+    pub fn new(root: &Path, file_extensions: Vec<String>, ignore_patterns: Vec<String>) -> Self {
         Self {
             root: root.to_path_buf(),
-            include_pattern,
-            ignore_patterns: default_ignore_patterns,
+            include_pattern: Self::construct_include_pattern(file_extensions),
+            ignore_patterns: Self::construct_ignore_patterns(ignore_patterns),
         }
     }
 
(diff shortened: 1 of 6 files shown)


crates/router/src/files.rs
```diff
@@ -17,110 +15,189 @@ pub const IGNORE_PATH_PREFIX: &str = "_";
...
-    fn should_ignore(&self, path: &Path) -> bool {
-        path.components().any(|c| match c {
-            Component::Normal(os_str) => {
-                (self.has_public && os_str == OsStr::new("public"))
-                    || os_str == OsStr::new(STORE_FOLDER)
-                    || os_str.to_string_lossy().starts_with(IGNORE_PATH_PREFIX)
-            }
-            _ => false,
-        })
+    fn construct_include_pattern(file_extensions: Vec<String>) -> String {
```
Angelmmiguel, line 54: The current codebase usually refers to these methods as `build_xxx`. This comment also applies to the `construct_ignore_pattern`. 

This is an important change, but a minor convention fix :)

crates/router/src/files.rs
```diff
@@ -17,110 +15,189 @@ pub const IGNORE_PATH_PREFIX: &str = "_";
...
+        for default_extension in Self::DEFAULT_EXTENSIONS {
+            file_extensions.push(default_extension.to_string());
+        }
+
+        format!("**/*.{{{}}}", file_extensions.join(","))
+    }
+
+    fn construct_ignore_patterns(ignore_patterns: Vec<String>) -> Vec<String> {
+        let mut result = vec![
+            "**/public/**".to_string(),
```
Angelmmiguel, line 65: I would move this pattern into a global. Just to avoid having fixed strings arounds. Wdyt?

crates/router/src/files.rs
```diff
@@ -17,110 +15,189 @@ pub const IGNORE_PATH_PREFIX: &str = "_";
...
-        let tests = [
-            ("public/index.js", true),
-            ("examples/public/index.js", true),
-            ("examples/public/other.js", true),
-            ("public.js", false),
-            ("examples/public.js", false),
-            ("./examples/public.js", false),
-            ("./examples/index.js", false),
-        ];
+    fn walk_default() {
```
Angelmmiguel, line 82: This is amazing! Thank you for completing all the tests with the different use cases 😄

crates/router/src/files.rs
```diff
@@ -17,110 +15,189 @@ pub const IGNORE_PATH_PREFIX: &str = "_";
...
+            actual.insert(PathBuf::from_slash(String::from(
+                entry.path().to_string_lossy(),
+            )));
+        }
+
+        assert_eq!(expected, actual);
+    }
+
+    #[test]
+    fn walk_ignore2() {
```
Angelmmiguel, line 180: I would make this test name a bit more informative. What's the different between this and the previous one?

Maybe something like `walk_ignore_multiple_patterns`?

src/main.rs
```diff
@@ -32,6 +32,10 @@ pub struct Args {
     #[arg(long, default_value = "")]
     prefix: String,
 
+    /// Patterns to ignore when looking for worker files
+    #[arg(long, default_value = "", value_delimiter=';', num_args= 0..)]
+    ignore: Vec<String>,
```
Angelmmiguel, line 37: I really like you included support for multiple ignore patterns. Regarding the current argument configuration, it requires you to understand you need to join multiple patterns using a specific delimiter.

I found a common practice in other tools to repeat the config paremeter to add new values: `wws --ignore "node_modules" --ignore "target" .`. I think this is the default configuration when you don't specified the `value_delimiter` nor `num_args`. [Here you can find an example](https://users.rust-lang.org/t/clap-is-it-possible-to-have-options-with-multiple-values-rather-than-multiple-repeats/82755).

Wdyt?

src/main.rs
```diff
@@ -102,9 +106,9 @@ async fn main() -> std::io::Result<()> {
         }
 
         println!("⚙️  Loading routes from: {}", &args.path.display());
-        let routes = Routes::new(&args.path, &args.prefix, &config);
+        let routes = Routes::new(&args.path, &args.prefix, args.ignore, &config);
 
-        println!("🗺  Detected routes:");
+        println!("🗺  Detected {} routes:", routes.routes.len());
```
Angelmmiguel, line 111: I like this change! We also have a similar requested change on #103, although that one is slightly different. `Routes` already initializes the Wasm modules. In #103, we expect to show this information before they get processed.

Angelmmiguel (commented):
The overall changes LGTM. The refactor and new tests are great 😄. I added a few minor comments.

Thank you!

ereslibre (approved):
LGTM besides @Angelmmiguel comments, neat contribution!

Angelmmiguel:
@assambar thank you for the update! Please, rebase the changes from `main` as you would need to configure the new source in the `deny.toml` file. Here you have some documentation about it: https://embarkstudios.github.io/cargo-deny/checks/sources/cfg.html.

Angelmmiguel (approved):
LGTM! Amazing contribution 😄

Issue #95, opened 2023-02-16
Add support for Golang


Angelmmiguel:
[Golang](https://go.dev/) is a popular programming language. Currently, you can compile Go apps to Wasm + WASI using [TinyGo](https://tinygo.org/docs/guides/webassembly/). The Go team will release a formal support for Wasm + WASI (preview1) in 1.21 (August 2023).

Golang can be compiled to WebAssembly, so there's no requirement of a language runtime like Python, Ruby or JavaScript. The goal for this task is to create the wrapper code to interact with Wasm Workers Server. The same approach we took for the [Rust support](https://github.com/vmware-labs/wasm-workers-server/tree/main/kits/rust).

## Technical requirements

These are the for the Golang support:

1. Read the JSON metadata that comes via STDIN. Here you have the example from the Rust kit:

    https://github.com/vmware-labs/wasm-workers-server/blob/5631fb400c03d3239daed03b523914b87717ad83/kits/rust/src/io.rs#L15-L23

1. Define the `Request`, `Response` and `Cache` entities. If it's available, it's better to use primitives from the Golang standard library or common usage libraries. 
1. Initialize a `Request` object from the JSON metadata.
1. Initialize the Cache store.
1. Call the worker function with the `Request` instance.
1. Detect the content in the response. If it's a valid UTF-8 string, we can return. If it uses a different encoding, you need to encode it in base64.
1. Transform the response and the Cache status into the expected JSON output. Note that the `base64` flag is only enabled when the content is encoded. 

    https://github.com/vmware-labs/wasm-workers-server/blob/5631fb400c03d3239daed03b523914b87717ad83/kits/rust/src/io.rs#L59-L65

1. Print the JSON output via STDOUT

## Additional notes

* The library should be defined in the `kits/go` folder.
* The documentation is available in the site: https://workers.wasmlabs.dev/docs/languages/introduction.
* Currently, we're not publishing the libraries to language registries. Instead, we document how to get it from the GitHub repository.

Angelmmiguel:
That's amazing @mnafees! I assigned the task to you to keep track of it. I also updated the task to provide a better description of the requirements to complete it.

Please, feel free to ask any question and open a draft PR in case you want to share some early code for feedback 😄

mnafees:
Hi @Angelmmiguel, opened a basic draft PR. I will be updating the PR with examples in a following commit!

Angelmmiguel:
[...]

PR #129, opened 2023-05-16
Add support for Go 

mnafees:
This PR adds support for Go support. 

Currently, only compiled WASM modules using TinyGo is supported. This comes as a result of TinyGo supporting WASI whereas the original Go compiler emits WASM that needs to be accompanied with a [wrapper JS file](https://github.com/golang/go/blob/master/misc/wasm/wasm_exec.js).

**To compile**
```bash
tinygo build -o my_worker.wasm -target wasi my_worker.go
```

Resolves https://github.com/vmware-labs/wasm-workers-server/issues/95


kits/go/worker/worker.go
```diff
@@ -0,0 +1,163 @@
...
+	kvOut := "{"
+
+	for k, v := range cache {
+		kvOut += fmt.Sprintf(`"%s":"%s",`, k, v)
+	}
+
+	kvOut = strings.TrimSuffix(kvOut, ",") + "}"
+
+	fmt.Printf("{\"data\":\"%s\",\"headers\":%s,\"status\":%d,\"kv\":%s,\"base64\":%t}",
+		strings.ReplaceAll(o.data, "\"", "\\\""), headersOut, o.status, kvOut, o.base64)
```
Angelmmiguel, line 81: For creating the JSON, did you consider using [the `SJSON` library](https://github.com/tidwall/sjson)? The current method works properly, although it seems more convenient for me to use a library that takes care of proper formatting.

I'm not a go expert, so if there's any caveat of using `SJSON`, feel free to keep the current approach.

kits/go/worker/worker.go
```diff
@@ -0,0 +1,163 @@
...
+
+	return len(o.data), nil
+}
+
+func (o *output) WriteHeader(statusCode int) {
+	o.status = uint16(statusCode)
+}
```
Angelmmiguel, line 88: I see this method set the status code for the request. Probably it's a leftover, but I believe we should call this method `SetStatus`.

Angelmmiguel (commented):
Hello @mnafees ,

Thank you for this amazing contribution! I added two minor comments. The code looks great and all the examples work perfectly.

Apart from the examples, we should include a [new documentation page for Go](https://github.com/vmware-labs/wasm-workers-server/tree/main/docs/docs/languages). It's not a blocker for merging this, so we can work on it in a separate PR. If it's fine for you, I would add a note in that documentation page mentioning you as the original author of the Go kit.

Thank you! 🎉

kits/go/worker/worker.go
mnafees: Since the current implementation works only when compiled with TinyGo, using Go's JSON standard library is unsupported because it uses reflection and TinyGo does not support reflection at the moment. The `SJSON` library itself seems to be using the standard Go JSON library which will fail to run as a WASI binary unfortunately.

This limitation shall hopefully be dealt with once the official Go compiler supports the WASI backend.

kits/go/worker/worker.go
mnafees: The `output` struct implements Go's [`http.ResponseWriter`](https://pkg.go.dev/net/http#ResponseWriter) interface and hence it is expected to call this method `WriteHeader`.

mnafees:
> Hello @mnafees ,
> 
> Thank you for this amazing contribution! I added two minor comments. The code looks great and all the examples work perfectly.
> 
> Apart from the examples, we should include a [new documentation page for Go](https://github.com/vmware-labs/wasm-workers-server/tree/main/docs/docs/languages). It's not a blocker for merging this, so we can work on it in a separate PR. If it's fine for you, I would add a note in that documentation page mentioning you as the original author of the Go kit.
> 
> Thank you! 🎉

I am going to add code comments as well as the docs for the new Go support kit and update the PR accordingly :)

kits/go/worker/worker.go
Angelmmiguel: Thank you for the clarification! Let's keep it until [Golang release the official WASI support](https://github.com/golang/go/issues/58141#issuecomment-1554055160). I think the release will be around in August :)

kits/go/worker/worker.go
Angelmmiguel: Thank you for the clarification!

kits/go/worker/worker.go
Angelmmiguel: @mnafees by default, `wws` should print any data sent to the `stderr` inside the workers:

https://github.com/vmware-labs/wasm-workers-server/blob/41cd17723c0d6e2b94f4cea18526ce491bff7b0e/crates/worker/src/lib.rs#L125-L133

If the worker sends data to `stderr`, it should be printed in the terminal. We changed this behavior in a recent commit (not released yet). In the new code, we're inheriting the `stderr` from the environment, so it should print the error too.

If the error comes from the `Wasm` execution, you're totally right that `wws` is hiding that error. Currently, we're returning a generic error:

https://github.com/vmware-labs/wasm-workers-server/blob/5631fb400c03d3239daed03b523914b87717ad83/crates/server/src/handlers/worker.rs#L72-L79

I'm going to create a new task to show that error in `stderr` as it's required for both development and production. Meanwhile, please modify the previous code with this change and that will print the error:

```rust
// In /crates/server/src/handlers/worker.rs
let (handler_result, handler_success) =
    match route
        .worker
        .run(&req, &body_str, store, vars, stderr_file.get_ref())
    {
        Ok(output) => (output, true),
        Err(err) => {
            eprintln!("Error output from the worker:\n{err}");
            (WasmOutput::failed(), false)
        }
    };
```

Angelmmiguel: New issue to expose the Wasm backtrace: https://github.com/vmware-labs/wasm-workers-server/issues/131

Angelmmiguel:
Hey @mnafees! 

This is just a check on the status of this PR :). I believe the current Go kit is in a great shape to included in the project. 

If you are more busy lately, I can do a final review, merge it and create an issue to work on the documentation in a separate PR. If you prefer to complete the documentation on this PR, feel free to add it too. 

No rush for it. I just wanted to to do a check since the latest comment was from a few days ago.

Thanks!

mnafees:
Hi @Angelmmiguel, apologies for being AWOL for a while. I got busy with other work. I am updating the PR with the docs. I also went ahead and chose to use the `sjson` library as suggested by you. Will update with a new comment once the PR is ready for your re-review. Thank you for the patience!

mnafees:
@Angelmmiguel I have updated the PR with all of my changes to the documentation :)

Angelmmiguel (approved):
The changes look great @mnafees! Thank you very much for this contribution 😄. I'm really happy to see that `wws` supports Go now 👏.

Btw, I'll write an article about a new release and I'll be sure to include the recent contributions like this one there 😄.

Angelmmiguel:
Btw, no worries about the issue on the CI. It's a flaky test I need to fix in `main` and it's not related to your changes. I will retry it after other builds finish.

gzurl:
Awesome contribution @mnafees! 🚀

PR #134, opened 2023-05-23
feat: add e2e tests to check different examples are working properly

Angelmmiguel:
Introduce a new set of e2e tests to confirm `wws` binary is working properly. Currently, we're testing the following cases:

- `examples/js-basic`
- `examples/js-json`
- `examples/js-params`
- `examples/rust-basic`
- `examples/rust-kv`

I also introduce a change in the CI to build the binary in different platforms during the PR too. The end goal is to ensure PRs don't break the basic features.

PR #136, opened 2023-05-24
fix: use open_ambient_dir instead of fs::File::open to load folders

### LINKED ISSUE #135: JavaScript workers are not properly working on Windows
### Describe the bug

After adding a new end-to-end test suite on #134, the CI failed for JavaScript workers on Windows. I could reproduce the issue on a Windows machine and I'm researching the root cause now.

### Reproduction steps

1. Download the latest version of `wws`
2. Download a basic JavaScript worker from the [examples folder](https://github.com/vmware-labs/wasm-workers-server/blob/5631fb400c03d3239daed03b523914b87717ad83/examples/js-basic/index.js)
3. Run it
4. Get a 503 error 

### Expected behavior

The workers returns the expected 200 response.

### Additional context

Rust workers are working properly.

Angelmmiguel:
This PR fixes the issue when opening folders in Windows environments. Before, `wws` was using `fs::File::open` to open the directories for building the `WasiContext`. This approach worked properly on both Linux and MacOS, but was failing on Windows.

Now, `wws` open the directories with the [`Dir::open_ambient_dir` method](https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/sync/struct.Dir.html#method.open_ambient_dir) as the official [Wasmtime CLI does](https://github.com/bytecodealliance/wasmtime/blob/5b93cdb84def5fa7cb3ff823bc80216f19c296c5/src/commands/run.rs#L378).

From now on, these kind of errors will raise an error in the e2e tests (#134).

It closes #135

### DIFF
--- a/crates/runtimes/src/modules/external.rs
+++ b/crates/runtimes/src/modules/external.rs
@@ -7,7 +7,7 @@ use std::{
     fs,
     path::{Path, PathBuf},
 };
-use wasmtime_wasi::{Dir, WasiCtxBuilder};
+use wasmtime_wasi::{ambient_authority, Dir, WasiCtxBuilder};
 use wws_runtimes_manager::metadata::Runtime as RuntimeMetadata;
 use wws_store::Store;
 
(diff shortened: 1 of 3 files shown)


ereslibre (approved):
Great finding! LGTM

gzurl (approved):
Awesome!! LGTM 👍

Issue #131, opened 2023-05-22
Expose Wasm errors in stderr


Angelmmiguel:
### Is your feature request related to a problem? Please describe.

When processing a request in a worker, failures may happen at two different levels:

* Recoverable errors in the worker module: the module recovers from the error. It may print relevant information to `stderr` and continue processing the request or providing an error response.
* Wasm failures: the Wasm module fails and the worker cannot recover. This may be cause by a panic in the worker, unreachable functions, or other issues. The module cannot provide a proper response and error message. `wws` returns the error (Wasm backtrace) to the handler.

In the first case, we're already printing any data sent to `stderr` from the worker. After 52076845fc079e4677b1a8dc90e93ef2916d8428, we're inheriting the `stderr` from the environment and configuring it in the WASI context. You can also configure the `stderr` when using `wws` as a library.

However, for the second case we're not printing any error and just providing a default error page:

https://github.com/vmware-labs/wasm-workers-server/blob/5631fb400c03d3239daed03b523914b87717ad83/crates/server/src/handlers/worker.rs#L72-L79

This error exposes the Wasm backtrace. This information is very relevant for debugging on both development and production.

### Describe the solution you'd like

Before sending a generic error, `wws` must print the Wasm backtrace in the terminal. The `stderr` is configurable, so the `stderr_file` (`Option<File>`) may contain a file descriptor that should be use. If it's `None`, you can directly call `eprintln!` with the proper message. 

There's a different task to improve how `wws` manages errors (#73)

### Describe alternatives you've considered

None

### Additional context

- https://github.com/vmware-labs/wasm-workers-server/pull/129#discussion_r1200220022

carrycooldude:
Will love to work on this issue @Angelmmiguel

carrycooldude:
Just need to ask @Angelmmiguel , I just have to add Error Msg right?

Angelmmiguel:
> Just need to ask @Angelmmiguel , I just have to add Error Msg right?

Exactly @carrycooldude. The goal is to capture the error from the `run` method and display it in the `stderr`. However, the `stderr` is configurable, so it may be a specific file to write or the default `stderr` from the process. For that, the handler defines an `Option<File>` variable called `stderr_file`. 

If that variable is present, you should write the data to that file descriptor. If it's not present, you can just print the error using `eprintln!`.

I forgot to add that to the issue, so I also updated the description accordingly.

Angelmmiguel:
Closed by #141 🎉

PR #141, opened 2023-05-26
Added error message and exposes WASM error message

carrycooldude:
Reference to this issue: https://github.com/vmware-labs/wasm-workers-server/issues/131

Before sending a generic error, wws must print the Wasm backtrace in the terminal. The stderr is configurable, so the stderr_file (Option<File>) may contain a file descriptor that should be used. If it's None, you can directly call eprintln! with the right message.

### DIFF
--- a/crates/server/src/handlers/worker.rs
+++ b/crates/server/src/handlers/worker.rs
@@ -75,7 +75,7 @@ pub async fn handle_worker(req: HttpRequest, body: Bytes) -> HttpResponse {
                 .run(&req, &body_str, store, vars, stderr_file.get_ref())
             {
                 Ok(output) => (output, true),
-                Err(_) => (WasmOutput::failed(), false),
+                Err(_) => (eprintln!(), false),
             };
 
         let mut builder = HttpResponse::build(
--- a/crates/server/src/handlers/worker.rs
+++ b/crates/server/src/handlers/worker.rs
@@ -8,7 +8,7 @@ use actix_web::{
     web::{Bytes, Data},
     HttpRequest, HttpResponse,
 };
-use std::{fs::File, sync::RwLock, io::Write};
+use std::{fs::File, io::Write, sync::RwLock};
 use wws_router::Routes;
 use wws_worker::io::WasmOutput;
 
@@ -68,7 +68,7 @@ pub async fn handle_worker(req: HttpRequest, body: Bytes) -> HttpResponse {
             }
             None => None,
         };
-        
+
         let (handler_result, handler_success) =
             match route
                 .worker
@@ -77,16 +77,18 @@ pub async fn handle_worker(req: HttpRequest, body: Bytes) -> HttpResponse {
                 Ok(output) => (output, true),
                 Err(error) => {
                     if let Some(stderr_file) = stderr_file.get_ref() {
-                        if let Ok(mut stderr_file) = stderr_file.try_clone(){
-                            stderr_file.write_all(error.to_string().as_bytes()).expect("Failed to write error to stderr_file");
-                        }
-                        else{
+                        if let Ok(mut stderr_file) = stderr_file.try_clone() {
+                            stderr_file
+                                .write_all(error.to_string().as_bytes())
+                                .expect("Failed to write error to stderr_file");
+                        } else {
                             eprintln!("{}", error);
                         }
                     } else {
                         eprintln!("{}", error);
                     }
-                    (WasmOutput::failed(), false)},
+                    (WasmOutput::failed(), false)
+                }
             };
 
         let mut builder = HttpResponse::build(
--- a/crates/server/src/handlers/worker.rs
+++ b/crates/server/src/handlers/worker.rs
@@ -8,7 +8,7 @@ use actix_web::{
     web::{Bytes, Data},
     HttpRequest, HttpResponse,
 };
-use std::{fs::File, sync::RwLock};
+use std::{fs::File, sync::RwLock, io::Write};
 use wws_router::Routes;
 use wws_worker::io::WasmOutput;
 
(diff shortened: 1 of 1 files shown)


crates/server/src/handlers/worker.rs
```diff
@@ -75,7 +75,9 @@ pub async fn handle_worker(req: HttpRequest, body: Bytes) -> HttpResponse {
                 .run(&req, &body_str, store, vars, stderr_file.get_ref())
             {
                 Ok(output) => (output, true),
-                Err(_) => (WasmOutput::failed(), false),
+                Err(e) => {
+                    eprintln!("{}", e);
+                    (WasmOutput::failed(), false)},
```
Angelmmiguel, line 80: The `eprintln!` method prints the given content to the `stderr` descriptor associated to the running process. This is perfect when running `wws` as a CLI, but developers may set a specific `stderr_file` when using it as a library.

For that reason, we need to cover that use case here too. The `Option<File>` is present in the `stderr_file`. In line `75`,  you have an example about how it gets a reference to this object: 

```
.run(&req, &body_str, store, vars, stderr_file.get_ref())
```

With that reference, you can check if the `Option` is `Some` or `None`. In case of `Some`, you would need to write the content to that specific file. You can do it using the `write_all` method, although first you need to `try_clone` it to get a mutable reference ([see this part of the code as an example](https://github.com/vmware-labs/wasm-workers-server/blob/790671deb92c5abf9cd05152bec76ca783b8e091/crates/worker/src/lib.rs#L84)). 

Note that in that case, we're raising any error up to the caller (with the `?`) operator. In this case, we need to manage those. If you get any error while writing to the given file, as a fallback print the two errors (writing the file and original error) to the stderr via `eprintln!`.

Angelmmiguel (commented):
Thank you very much for your contribution @carrycooldude! 

I added some notes about managing the case when the `stderr` is configured to be printed in a specific file instead of the `stderr` descriptor from the running process 😄

carrycooldude:
Screenshot from 2023-05-26 19-19-51
Can you tell me why this error is coming? @Angelmmiguel

Angelmmiguel:
> Screenshot from 2023-05-26 19-19-51 Can you tell me why this error is coming? @Angelmmiguel

In this case, the error is related to the `stderr_file` type. Since this variable comes from actix `Data`, the type is `Data<Option<File>>` instead of `Option<File>`. You can get a reference to the inner element by calling the `stderr_file.get_ref()` method. That will return an `&Option<File>` that you can use 😄.

Also, I see you're using `expect` to ensure the functions return an `Ok` variable. In this case, this will throw a `panic` and won't return any proper result to the users. To avoid it and returning an error page to the user, you can default to `eprintln!` in case any of these functions fail.

crates/server/src/handlers/worker.rs
carrycooldude: > > Screenshot from 2023-05-26 19-19-51 Can you tell me why this error is coming? @Angelmmiguel
> 
> In this case, the error is related to the `stderr_file` type. Since this variable comes from actix `Data`, the type is `Data<Option<File>>` instead of `Option<File>`. You can get a reference to the inner element by calling the `stderr_file.get_ref()` method. That will return an `&Option<File>` that you can use smile.
> 
> Also, I see you're using `expect` to ensure the functions return an `Ok` variable. In this case, this will throw a `panic` and won't return any proper result to the users. To avoid it and returning an error page to the user, you can default to `eprintln!` in case any of these functions fail.



> > Screenshot from 2023-05-26 19-19-51 Can you tell me why this error is coming? @Angelmmiguel
> 
> In this case, the error is related to the `stderr_file` type. Since this variable comes from actix `Data`, the type is `Data<Option<File>>` instead of `Option<File>`. You can get a reference to the inner element by calling the `stderr_file.get_ref()` method. That will return an `&Option<File>` that you can use smile.
> 
> Also, I see you're using `expect` to ensure the functions return an `Ok` variable. In this case, this will throw a `panic` and won't return any proper result to the users. To avoid it and returning an error page to the user, you can default to `eprintln!` in case any of these functions fail.

I tried that too but same result , gave a reference to stderr_file

carrycooldude:
> > Screenshot from 2023-05-26 19-19-51 Can you tell me why this error is coming? @Angelmmiguel
> 
> In this case, the error is related to the `stderr_file` type. Since this variable comes from actix `Data`, the type is `Data<Option<File>>` instead of `Option<File>`. You can get a reference to the inner element by calling the `stderr_file.get_ref()` method. That will return an `&Option<File>` that you can use smile.
> 
> Also, I see you're using `expect` to ensure the functions return an `Ok` variable. In this case, this will throw a `panic` and won't return any proper result to the users. To avoid it and returning an error page to the user, you can default to `eprintln!` in case any of these functions fail.

Did that too also , 
```
let (handler_result, handler_success) =
            match route
                .worker
                .run(&req, &body_str, store, vars, &*stderr_file.get_ref())
            {
                Ok(output) => (output, true),
                Err(error) => {
                    if let Some(&stderr_file) = stderr_file {
                        let mut stderr_file = stderr_file.try_clone().expect("Failed to clone stderr_file");
                        stderr_file.write_all(error.to_string().as_bytes()).expect("Failed to write error to stderr_file");
                    } else {
                        eprintln!("{}", error);
                    }
                    (WasmOutput::failed(), false)},
            };
```

also, try to change to eprintln!() but getting error on that @Angelmmiguel

carrycooldude:
hey @Angelmmiguel , As checks are cleared of this PR, can you merge this?

Angelmmiguel (approved):
Changes LGTM! Thank you very much for your contribution @carrycooldude 😄👏

carrycooldude:
Thank You So Much for the support @Angelmmiguel , I will learn from you more 🙏
Let me know if I will do more issues like these

PR #143, opened 2023-05-29
feat: allow running JavaScript workers with async methods

### LINKED ISSUE #142: Support async methods in the JavaScript kit

Currently, the JavaScript workers don't allow you to use `async` methods for handlers. In some cases, this is convenient as the methods may need to wait for some resources before returning the response. `Async` methods are supported since Javy 1.0.0, so we should be able to implement them in `wws`.

This is an example of an `async` worker:

```js
// Using an async function to reply!
async function handle(request) {
  // Body response
  const body = `Hello world!`;

  // Build a new response
  let response = new Response(body);

  // Add a new header
  response.headers.set("x-generated-by", "wasm-workers-server");

  return response;
}

addEventListener('fetch', event => {
  event.respondWith(handle(event.request));
});
```

Even though it doesn't require an `async` method, we can use it as an example.

I would like to use `async` when defining methods in JavaScript workers. Some libraries and several examples that I found require this feature.

It's available in Javy 1.0.0, so it may require to upgrade the JavaScript kit and start using the new crate from the Javy team: https://crates.io/crates/javy.

_No response_

### Additional context

_No response_

Angelmmiguel:
Add support for async methods in JavaScript workers. For that, I split the JavaScript runtime in two steps:

* Call the `requestToHandler` method to get the `response` method / object.
* Convert `response` to a `Promise` so we cover all use cases (async / sync).
* Save the final response in a new `result` variable.

To add support on the JS Runtime side, I upgraded the JS kit to use the [new javy crate](https://github.com/bytecodealliance/javy/tree/main/crates/javy). It adds new features and a better high-level APIs for running JS code in Rust.

I also included a new example and a new e2e test.

It closes #142


ereslibre (approved):
LGTM 🎉; this is great!, although I think it would be good to split the main commit into two:

- quickjs -> javy, binary removal
- async feature

Angelmmiguel:
> LGTM 🎉; this is great!, although I think it would be good to split the main commit into two:
> 
> * quickjs -> javy, binary removal
> * async feature

Sure, let me do it for traceability on the PR, although they will be squashed on `main`.

Angelmmiguel:
Done @ereslibre ! 😄

ereslibre:
👏 Could we merge without squash? I think we can rebase and merge, no?

If so, that would be great! Thanks a lot @Angelmmiguel.

Angelmmiguel:
> 👏 Could we merge without squash? I think we can rebase and merge, no?
> 
> If so, that would be great! Thanks a lot @Angelmmiguel.

Yeah, that's possible too. However, I usually converge commits into a single one that refers to the PR. Sometimes PRs may include multiple commits like `fix: fmt` and `fix: typo`. Instead of asking people to rewrite the git history, I believe it's simpler to limit the PR scope and squash it.

ereslibre:
> Instead of asking people to rewrite the git history, I believe it's simpler to limit the PR scope and squash it.

👍, however if I'm not mistaken the GitHub option "Rebase and merge" does not enforce anybody to rebase their own history, given they will just fast forward on top these changes.

Anyhow, it's fine by me.

Angelmmiguel:
> > Instead of asking people to rewrite the git history, I believe it's simpler to limit the PR scope and squash it.
> 
> 👍, however if I'm not mistaken the GitHub option "Rebase and merge" does not enforce anybody to rebase their own history, given they will just fast forward on top of these changes.
> 
> Anyhow, it's fine by me.

Yeah, that option is available. I usually `squash` all PRs, but I agree it's interesting to keep the commits separated in `main` for this specific one. I will rebase and merge the two commits in this one 😄

PR #152, opened 2023-06-02
feat: allow git remote repositories (https) as project locations

### LINKED ISSUE #147: Run a project from a remote repository

There are many different ways a project / workers could be stored. Currently, we are allowing only local resources by pointing `wws` to a system folder:

```
wws ./my-folder
```

However, this restriction introduces an extra step of downloading and configuring the project in the environment. `wws` can identify the resources based on the path, extension and protocol, and prepare the environment on behalf of the user.

All the examples require you to clone / download workers. Ideally, the demos must be available in a single command:

```
wws https://github.com/vmware-labs/wasm-workers-server.git
```

`wws` must detect the resource and prepare the environment based on it. For this initial implementation, it will detect links that ends as `.git` to process them as `git` repositories. I would like to set other specific configurations related to git like:

* `--git-folder`
* `--git-branch`
* `--git-rev` 

The current alternative is to ask you to clone the repo in advance. We want to avoid this and simplify the process.

### Additional context

Related issue from @k33g (#121).

Angelmmiguel:
Add support for git repositories as valid project locations. Now you can run a project directly from a remote repository. There are different options to configure the git repository. These are all the new options:

```
--git-commit <GIT_COMMIT>  Set the commit when using a git repository as project
--git-tag <GIT_TAG>        Set the tag when using a git repository as project
--git-branch <GIT_BRANCH>  Set the branch when using a git repository as project
--git-folder <GIT_FOLDER>  Change the directory when using a git repository as project
```

Here you have an example:

```
$ cargo run -- https://github.com/vmware-labs/wasm-workers-server.git --git-folder "examples/js-basic"
⚙️  Loading routes from: /tmp/dd21e3cd6d0f515301e1c7070e562af06074d9e8d10566179f97dba47e74cec9/examples/js-basic
✅ Workers loaded in 2.113770416s.
    - http://127.0.0.1:8080/
      => /tmp/dd21e3cd6d0f515301e1c7070e562af06074d9e8d10566179f97dba47e74cec9/examples/js-basic/index.js
```

It closes #147

## Install missing language runtimes

Some workers require specific runtimes. For example, you need to download a Python or Ruby runtime to develop workers based on these languages. `wws` stores this information in the `.wws.toml`, so you can install them with `wws runtimes install`. 

When pulling a repository is convenient to install the missing runtimes directly. However, we may not want to make this the default behavior as you need to download "arbitrary" binaries for Internet. For this reason, I added an option to install runtimes automatically, so you can enable it for projects you trust:

```
$ wws https://github.com/vmware-labs/wasm-workers-server.git --git-folder "examples/js-basic" --install-runtimes
$ # Or
$ wws https://github.com/vmware-labs/wasm-workers-server.git --git-folder "examples/js-basic" -i
```

### DIFF
--- a/crates/project/src/types/git.rs
+++ b/crates/project/src/types/git.rs
@@ -1,54 +1,66 @@
 // Copyright 2023 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::options::Options;
-use anyhow::{bail, Result};
+use crate::options::{GitReference, Options};
+use anyhow::{anyhow, bail, Result};
 use git2::{Oid, Repository};
 use sha256::digest as sha256_digest;
-use std::{env::temp_dir, fs::remove_dir_all, path::PathBuf};
+use std::{
+    env::temp_dir,
+    fs::remove_dir_all,
+    path::{Path, PathBuf},
+};
 
 // Default remote for git repos
 static DEFAULT_REMOTE: &str = "origin";
 
 /// Prepare a project based on a git repository. This method
 /// clones the repo locally and returns the path in which it's located.
-pub fn prepare_git_project(location: &str, options: Option<Options>) -> Result<PathBuf> {
+pub fn prepare_git_project(location: &Path, options: Option<Options>) -> Result<PathBuf> {
+    let project_url = location
+        .to_str()
+        .ok_or(anyhow!("The project URL cannot be retrieved"))?;
     // By default, we use temporary dirs
-    let mut dir = temp_dir().join(sha256_digest(location));
+    let mut dir = temp_dir().join(sha256_digest(project_url));
 
     if dir.exists() {
         // Clean up a previous download
         remove_dir_all(&dir)?;
     }
 
-    let repo = match Repository::clone(location, &dir) {
+    let repo = match Repository::clone(project_url, &dir) {
         Ok(repo) => repo,
         Err(e) => bail!("There was an error cloning the repository: {e}"),
     };
 
     if let Some(options) = options {
         if let Some(git) = options.git {
-            // These options are prioritized
-            if let Some(commit) = git.commit {
-                let oid = Oid::from_str(&commit)?;
-                let commit = repo.find_commit(oid)?;
-                repo.checkout_tree(commit.as_object(), None)?;
-            } else if let Some(tag) = git.tag {
-                let mut remote = repo.find_remote(DEFAULT_REMOTE)?;
-                let tag_remote = format!("refs/tags/{tag}:refs/tags/{tag}");
-                remote.fetch(&[&tag_remote], None, None)?;
+            if let Some(git_ref) = git.git_ref {
+                match git_ref {
+                    GitReference::Commit(commit) => {
+                        let oid = Oid::from_str(&commit)?;
+                        let commit = repo.find_commit(oid)?;
+                        repo.checkout_tree(commit.as_object(), None)?;
+                    }
+                    GitReference::Tag(tag) => {
+                        let mut remote = repo.find_remote(DEFAULT_REMOTE)?;
+                        let tag_remote = format!("refs/tags/{tag}:refs/tags/{tag}");
+                        remote.fetch(&[&tag_remote], None, None)?;
 
-                let oid = Oid::from_str(&tag)?;
-                let tag = repo.find_tag(oid)?;
-                repo.checkout_tree(tag.as_object(), None)?;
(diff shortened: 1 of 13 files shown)


src/main.rs
```diff
@@ -100,13 +156,25 @@ async fn main() -> std::io::Result<()> {
...
+                    Err(err) => {
+                        eprintln!("❌ There was an error installing the missing runtimes.");
+                        eprintln!("❌ Here you can find more information: {err}.");
+
+                        exit(1);
+                    }
+                }
+            } else {
+                println!("⚠️  Required language runtimes are not installed. Some files may not be considered workers");
+                println!("⚠️  You can install the missing runtimes adding the --install-runtimes / -i flag");
```
ereslibre, line 172: UX: being wws a close-to-the-developer tool, I think it would be better to install the missing runtime by default, so that the run without arguments does the expected thing.

I would add a `--do-not-install-runtimes` or something similar, for security minded folks that don't want the automatic runtime download to happen.

src/utils/runtimes.rs
```diff
@@ -0,0 +1,102 @@
...
+/// Default repository name
+pub const DEFAULT_REPO_NAME: &str = "wasmlabs";
+/// Default repository URL
+pub const DEFAULT_REPO_URL: &str = "https://workers.wasmlabs.dev/repository/v1/index.toml";
+
+/// Environment variable to set the repository name
+pub const WWS_REPO_NAME: &str = "WWS_REPO_NAME";
+pub const WWS_REPO_URL: &str = "WWS_REPO_URL";
+
+/// Loads the local configuration and install any missing runtime from it.
```
ereslibre, line 20:
```suggestion
/// Loads the local configuration and installs any missing runtime from it.
```

src/utils/runtimes.rs
```diff
@@ -0,0 +1,102 @@
...
+                install_runtime(project_root, &repo.name, runtime).await?;
+            }
+        }
+    }
+
+    println!("✅ Done");
+    Ok(())
+}
+
+/// Retrieves the remote repository and install the desired runtime.
```
ereslibre, line 46:
```suggestion
/// Retrieves the remote repository and installs the desired runtime.
```

ereslibre (commented):
Looks great. Very nice feature! LGTM, some comments from my side :)

src/main.rs
```diff
@@ -24,9 +29,9 @@ pub struct Args {
     #[arg(short, long, default_value_t = 8080)]
     port: u16,
 
-    /// Folder to read WebAssembly modules from
+    /// Location of the wws project. It could be a local folder or a git repository.
     #[arg(value_parser, default_value = ".")]
-    path: PathBuf,
+    location: String,
```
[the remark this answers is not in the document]
Angelmmiguel: Oh! I assumed that git repositories like `https://github.com/vmware-labs/wasm-workers-server` won't be valid paths. I should be checked it first. I'm going to revert these changes.

src/main.rs
Angelmmiguel: Currently, `wws` focuses on development environments, but in the future we will increase this scope. Since we don't have yet a `dev` command, the CLI cannot assume it's on a development environment.

I would enable `-i` flag by default when running it on development. For this initial release, I would keep it disabled by default. We can change it later and even adding an environment variable that enables it by default 😄

src/main.rs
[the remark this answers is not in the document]
Angelmmiguel: Fixed! I reverted these changes and use `PathBuf` instead.

crates/project/src/types/git.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    let repo = match Repository::clone(project_url, &dir) {
+        Ok(repo) => repo,
+        Err(e) => bail!("There was an error cloning the repository: {e}"),
+    };
+
+    if let Some(options) = options {
+        if let Some(git) = options.git {
+            if let Some(git_ref) = git.git_ref {
```
ereslibre, line 38: I have been looking for a better way to express this, but it requires [`let_chains`](https://github.com/rust-lang/rust/issues/53667) and is unstable. :(

ereslibre (approved):
LGTM, thanks @Angelmmiguel!

crates/project/src/types/git.rs
Angelmmiguel: Yeah, a new approach for "let chains" would be really great 😄. I will keep an eye on this. Thanks for the pointer.

Issue #153, opened 2023-06-06
Local panel to manage the current project


Angelmmiguel:
### Is your feature request related to a problem? Please describe.

After running `wws`, the only information you get is the output from the CLI. This information is limited to the loaded workers and you lose track of how they run over time. For example, you need to rely on workers configuration to identify if a worker has an associated K/V store or not.

In the future, `wws` will include more and more services making it difficult to track this information. Adding more data to the CLI output is possible, although it may end up being confusing.  For this reason, a read-only local panel could be a great addition to visualize the information properly.

Looking towards v2.0.0, `wws` should improve the process of deploying your projects into production. This read-only local panel is the seed for a more complete panel + API to deploy services from your local machine to a self-hosted `wws` service.

#### Architecture

This is a high level architecture about how this feature works. `wws` exposes a management API that will be consumed by both panel and new CLI commands

```mermaid
---
title: Wasm Workers Server
---
flowchart TD
    api(Management API)
    panel(Panel)
    
    wws -- exposes --- api
    wws -- exposes --- panel
    panel -- uses --> api

    CLI -- uses --> api
```

#### Initial version

This task, we will scope the panel to a read-only one. It will be disabled by default, so you can enable it via `--enable-panel` flag.

### Describe the solution you'd like

Provide a local panel in which I can get information about the running workers as well as required features. In the future, it will include more information such stats.

Regarding the management API, it only exposes the required information. In the future, we will add more features and a full read/write API for managing projects locally and in a remote `wws` deployment.

### Tasks

- [x] Create the management API (via #154)
- [ ] Add a basic administration panel

PR #154, opened 2023-06-12
feat: add the management API (readonly). Generate the OpenAPI spec

Angelmmiguel:
_All the context is available in #153_

Introduce a management API for Wasm Workers Server. For now, this API has only two endpoints:

- `/_api/v0/workers`: returns all the workers in the current project.
- `/_api/v0/workers/{id}`: returns the details of a worker like its configuration.

Currently, workers doesn't have any associated identifier (ID). I'm generating it by computing the SHA256 of the `{project_root}/{file}` path. 

### API Implementation

The API is split into two crates:

- `api-manage`: contains the source code of the API. It includes the models and the handlers. It exposes a method to configure an app service with these new methods. By default, `wws` doesn't enable the management API.
- `api-manage-openapi`: renders the OpenAPI specification from the `api-manage` and expose it. We will reuse this crate in the future to generate the CLI client in Rust (See #153).

To define the OpenAPI spec, I'm using [utoipa](https://github.com/juhaku/utoipa), an amazing crate to generate OpenAPI specifications from rust web frameworks like Actix. I only need to add annotations to the methods and models and it generates all the information.

Note that `utoipa` exposes the OpenAPI spec as an HTTP endpoint, but we don't want to expose it for now. That's the reason we're using the `api-manage-openapi` spec crate to generate it into a file. The process is done in the [`build.rs` file of that crate](https://github.com/vmware-labs/wasm-workers-server/blob/a83f36996e1e02485479693bf2e0a73024fb4cfb/crates/api-manage-openapi/build.rs).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,7 +88,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "465a6172cf69b960917811022d8f29bc0b7fa1398bc4f78b3c466673db1213b6"
 dependencies = [
  "quote",
- "syn",
+ "syn 1.0.107",
 ]
 
 [[package]]
@@ -203,7 +203,7 @@ dependencies = [
  "actix-router",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.107",
 ]
 
 [[package]]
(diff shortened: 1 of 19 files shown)


ereslibre (approved):
Super! LGTM 👏

PR #156, opened 2023-06-12
feat: add the initial administration panel

### LINKED ISSUE #153: Local panel to manage the current project

Angelmmiguel:
As Wasm Workers Server is evolving, we want to start adding an UI to start managing it. I'm really happy to open this PR as an initial version of that UI 😄. This admin panel is enabled via the `--enable-panel` flag and the UI looks like this:

```
$ wws https://github.com/vmware-labs/wasm-workers-server.git --git-folder "examples/python-basic" -i --enable-panel
⚙️  Checking local configuration...
🚀 Installing: wasmlabs - python / 3.11.1+20230217
✅ Done
⚙️  Loading routes from: /tmp/dd21e3cd6d0f515301e1c7070e562af06074d9e8d10566179f97dba47e74cec9/examples/python-basic
✅ Workers loaded in 489.699125ms.
    - http://127.0.0.1:8080/
      => /tmp/dd21e3cd6d0f515301e1c7070e562af06074d9e8d10566179f97dba47e74cec9/examples/python-basic/index.py
🎛️  The admin panel is available at http://127.0.0.1:8080/_panel/
```
Home page showing a card with the number of current workers
List all the current workers in the server
Show the details of a specific worker

In the future, we will start adding more features to it.

### Development

Currently, the UI is shipped as a crate (panel) and integrated in the main project. The client uses [React](https://react.dev/), [Clarity Core](https://core.clarity.design/) and [Vite](vitejs.dev/). To develop this UI you need to install NodeJS and NPM in your environment.

Cargo automatically builds the UI based from the build.rs file. If `npm` is not available, the build pipeline will success, although the panel won't be available. We designed this way to avoid breaking build pipelines that don't relate to the client when `node` / `npm` is not available.

### Integration

To integrate the compiled admin panel into the project, I used the [rust-embed](https://github.com/pyrossh/rust-embed) project. It automatically picks the files in a folder, embed them in the binary and provide an API to get the file content. With this approach we keep the server as a single-binary that it's easier to distribute.

It closes #153

### DIFF
--- a/crates/panel/build.rs
+++ b/crates/panel/build.rs
@@ -1,0 +1,36 @@
+use std::{io::ErrorKind, process::Command};
+
+// Build the client admin panel.
+fn main() {
+    // First check if NPM is available in the system
+    match Command::new("npm").spawn() {
+        Ok(_) => {
+            Command::new("npm")
+                .current_dir("client")
+                .arg("install")
+                .status()
+                .expect("failed to execute process");
+
+            Command::new("npm")
+                .current_dir("client")
+                .args(["run", "build"])
+                .status()
+                .expect("failed to execute process");
+        }
+        Err(e) => {
+            if let ErrorKind::NotFound = e.kind() {
+                eprintln!("`npm` was not found in your system. Please, install NodeJS / NPM to build the admin panel.");
+                eprintln!("See: https://nodejs.dev/en/download/");
+            } else {
+                eprintln!(
+                    "There was an error when building the admin panel with NodeJS / NPM: {e}"
+                );
+            }
+        }
+    }
+
+    // Tell Cargo that if the given file changes, to rerun this build script.
+    println!("cargo:rerun-if-changed=client/src/*");
+    println!("cargo:rerun-if-changed=client/index.html");
+    println!("cargo:rerun-if-changed=client/vite.config.js");
+}
--- a/crates/panel/build.rs
+++ b/crates/panel/build.rs
@@ -31,6 +31,7 @@ fn main() {
 
     // Tell Cargo that if the given file changes, to rerun this build script.
     println!("cargo:rerun-if-changed=client/src/*");
+    println!("cargo:rerun-if-changed=client/public/*");
     println!("cargo:rerun-if-changed=client/index.html");
     println!("cargo:rerun-if-changed=client/vite.config.js");
 }
--- a/crates/panel/client/src/components/sidebar.scss
+++ b/crates/panel/client/src/components/sidebar.scss
@@ -1,0 +1,26 @@
+// Copyright 2023 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+.sidebar {
+    display: flex;
+    flex-direction: column;
+    // 71px from the header
+    min-height: calc(100vh - 71px);
+    position: relative;
+    width: 12rem;
+
+    // Add a border with after due to collision with
+    // the nav items.
+    &:after {
+        content: "";
+        background-color: var(--cds-alias-status-disabled-tint);
+        height: 100%;
+        width: 1px;
+        position: absolute;
+        right: -1px;
+    }
+
+    a {
+        text-decoration: none;
+    }
+}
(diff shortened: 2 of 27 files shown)


crates/panel/client/src/components/sidebar.scss
```diff
@@ -0,0 +1,26 @@
+// Copyright 2023 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+.sidebar {
+    display: flex;
+    flex-direction: column;
+    // 71px from the header
+    min-height: calc(100vh - 71px);
```
ereslibre, line 8: It has been a while since my last interaction with CSS, I guess it's fine to use hardcoded values like this? Can we use a variable or something similar?

crates/panel/client/src/routes/home.jsx
```diff
@@ -0,0 +1,54 @@
...
+    fetch("/_api/v0/workers")
+      .then(res => res.json())
+      .then(json => setResult(json));
+  }, []);
+
+  return <>
+    <h2 cds-text="heading">Server</h2>
+    {result === undefined ? (
+      <CdsProgressCircle size="xl" />
+    ) : (
```
ereslibre, line 29: It is nice to show a progress indicator this way, by initializing the state to undefined and checking it here, until it has some data back from the server and we fall to the other part of the conditional 👍

ereslibre (approved):
LGTM, thanks @Angelmmiguel! A non-blocking minor question.

crates/panel/build.rs
```diff
@@ -0,0 +1,36 @@
...
+                    "There was an error when building the admin panel with NodeJS / NPM: {e}"
+                );
+            }
+        }
+    }
+
+    // Tell Cargo that if the given file changes, to rerun this build script.
+    println!("cargo:rerun-if-changed=client/src/*");
+    println!("cargo:rerun-if-changed=client/index.html");
+    println!("cargo:rerun-if-changed=client/vite.config.js");
```
assambar, line 35: I assume `client/public` is for the public assets. If that's the case shouldn't you add a `rerun-if-changed` for that folder, too?

assambar (approved):
Looks great!

crates/panel/client/src/components/sidebar.scss
Angelmmiguel: For now, I kept it fixed to the header size. It would be better to use a more dynamic value, but it requires a static one with the current layout. I will explore other options, although it requires to change the layout so I would do it in a separate PR.

crates/panel/build.rs
Angelmmiguel: Good point! Let me add that folder too.

Angelmmiguel: Done! 👍

PR #158, opened 2023-06-19
feat: reuse a git repository when it was previously cloned

### LINKED ISSUE #155: Reuse existing git repositories

When using Wasm Workers Server with a remote git repository, I always have to clone it entirely. For some large repositories, cloning may take long time. Even with this repository, it takes some seconds to clone it:

```
$ wws https://github.com/vmware-labs/wasm-workers-server.git \
  --git-folder "examples/js-basic" -i
```

I want `wws` to reuse the previously cloned repository instead of pulling it every time I run the command. To avoid having a stale repository, it should pull the latest changes. There's a [pull example in the git2-rs repository](https://github.com/rust-lang/git2-rs/blob/master/examples/pull.rs).

For now, a fast-forward approach should be enough. If the repository is "dirty" and the merge fails, you will need to delete the project folder and clone it again. You can always do the merge manually. 

The alternative is to keep the current approach. We may end up with conflicts, but for most cases this improvement is very convenient.

### Additional context

_No response_

Angelmmiguel:
Before, if `wws` detected that a repository was already cloned, it deleted the folder and clone it again. This PR changes this behavior. Now, it opens the repository and updates it to the given reference. 

By default, it tries to fetch the changes in the default branch (`main` / `master`). If you provide a git reference, it fetches the data from the origin. In case you provide a `commit`, it pulls the default branch too.

It closes #155

### DIFF
--- a/crates/project/src/types/git.rs
+++ b/crates/project/src/types/git.rs
@@ -25,17 +25,8 @@ pub fn prepare_git_project(location: &Path, options: Options) -> Result<PathBuf>
 
     let repo = if dir.exists() {
         // Reuse the same repository.
-        let repo = Repository::open(&dir)
-            .map_err(|e| anyhow!("There was an error opening the repository: {e}"))?;
-
-        // Pull latest changes
-        if let Some(GitReference::Branch(branch)) = git_ref.as_ref() {
-            pull_repository(&repo, branch)?;
-        } else {
-            pull_default_branch(&repo)?;
-        }
-
-        repo
+        Repository::open(&dir)
+            .map_err(|e| anyhow!("There was an error opening the repository: {e}"))?
     } else {
         // clone it
         Repository::clone(project_url, &dir)
@@ -45,6 +36,8 @@ pub fn prepare_git_project(location: &Path, options: Options) -> Result<PathBuf>
     if let Some(git_ref) = git_ref.as_ref() {
         match git_ref {
             GitReference::Commit(commit) => {
+                pull_default_branch(&repo)?;
+
                 let oid = Oid::from_str(commit)?;
                 repo.set_head_detached(oid)?;
                 repo.checkout_head(Some(&mut default_checkout()))?;
@@ -66,6 +59,8 @@ pub fn prepare_git_project(location: &Path, options: Options) -> Result<PathBuf>
                 repo.checkout_head(Some(&mut default_checkout()))?;
             }
         }
+    } else {
+        pull_default_branch(&repo)?;
     }
 
     if let Some(folder) = folder {
--- a/crates/project/src/types/git.rs
+++ b/crates/project/src/types/git.rs
@@ -3,11 +3,10 @@
 
 use crate::options::{GitReference, Options};
 use anyhow::{anyhow, bail, Result};
-use git2::{Oid, Repository};
+use git2::{build::CheckoutBuilder, FetchOptions, Oid, Repository};
 use sha256::digest as sha256_digest;
 use std::{
     env::temp_dir,
-    fs::remove_dir_all,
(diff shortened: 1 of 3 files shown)


crates/project/src/types/git.rs
```diff
@@ -16,58 +15,136 @@ static DEFAULT_REMOTE: &str = "origin";
...
+    // For now, we only distinguish between the two most common branch names.
+    // Ask the user to set the branch in any other case.
+    if repo.find_branch("main", git2::BranchType::Local).is_ok() {
+        Ok("main")
+    } else if repo.find_branch("master", git2::BranchType::Local).is_ok() {
+        Ok("master")
+    } else {
+        bail!("Couldn't find the default main branch. Please, set the Git branch you want to use.")
+    }
+}
```
ereslibre, line 111: Have been trying to see if there is a way with the git2 crate to find the default branch, I think it should be possible but couldn't find it at first glance -- in any case if we default to a shallow or blobless clone I'm not sure if that information will be available.

We can always improve it if necessary. There are projects that default their branch to, say, `release-X.Y` branch. There's the manual flag for that too, so we are good.

ereslibre (approved):
LGTM

crates/project/src/types/git.rs
```diff
@@ -16,58 +15,136 @@ static DEFAULT_REMOTE: &str = "origin";
...
-        remove_dir_all(&dir)?;
-    }
-
-    let repo = match Repository::clone(project_url, &dir) {
-        Ok(repo) => repo,
-        Err(e) => bail!("There was an error cloning the repository: {e}"),
+    let repo = if dir.exists() {
+        // Reuse the same repository.
+        Repository::open(&dir)
+            .map_err(|e| anyhow!("There was an error opening the repository: {e}"))?
```
assambar, line 29: Maybe add the path to the folder in the error message?
So people can find it and purge it in case of errors.

crates/project/src/types/git.rs
```diff
@@ -16,58 +15,136 @@ static DEFAULT_REMOTE: &str = "origin";
...
+    } else {
+        // clone it
+        Repository::clone(project_url, &dir)
+            .map_err(|e| anyhow!("There was an error cloning the repository: {e}"))?
```
assambar, line 33: I would add the path, here too. For same reasons

assambar (approved):
Ship it!

crates/project/src/types/git.rs
Angelmmiguel: I was looking at this too, but I couldn't find that method neither :/. So, I decided to detect the most common branches or ask the user to specify it.

Regarding blobless mode, I would love to use it but it's not available in the git2-rs / libgit2 library. The shallow support [was added to libgit2 a few weeks ago](https://github.com/libgit2/libgit2/issues/3058#issuecomment-1540940627) and it's not in git2-rs yet.

crates/project/src/types/git.rs
Angelmmiguel: The folder is printed in a previous message, that's the reason I'm not showing it here :)

```
⚙️  Loading routes from: /tmp/dd21e3cd6d0f515301e1c7070e562af06074d9e8d10566179f97dba47e74cec9/examples/js-basic
✅ Workers loaded in 108.82825ms.
    - http://127.0.0.1:8080/
      => /tmp/dd21e3cd6d0f515301e1c7070e562af06074d9e8d10566179f97dba47e74cec9/examples/js-basic/index.js
```

crates/project/src/types/git.rs
Angelmmiguel: Same comment, it's already printed previously.

PR #160, opened 2023-06-20
docs: use remote repos in examples and docs. Add a README.md to examples

### LINKED ISSUE #159: Update the documentation to use remote repositories for examples

Now that we support git repository URLs as a project location, we can simplify all the examples and the documentation. Instead of asking to download or clone a file, we can point directly to the git repository location and the right folder.

All examples points to the repository so we remove one extra step to run and check `wws` examples. These are the tasks to complete:

- [ ] Update documentation to use the new git repository feature.
- [ ] Add a README.md file to every example folder adding a simple description, a table, related documentation and a single command.

_No response_

### Additional context

_No response_

Angelmmiguel:
Update the documentation to take advantage of the remote locations features. Now, developers can try `wws` without having to clone or download a worker.

In addition to that, I crated a README.md file per every example, so any developer looking at them can see how easy is to run it.

It closes #159


ereslibre (approved):
LGTM, I think all new README.md are great 👍

Issue #132, opened 2023-05-23
Bump Wasmtime to 10.0.0 and other dependencies


Angelmmiguel:
### Is your feature request related to a problem? Please describe.

Currently, Wasm Workers Server uses Wasmtime 6. There are several improvements in newer versions so we should upgrade it to the latest version.

### Describe the solution you'd like

Wasm Workers Server dependencies are up-to-date and the compiled binaries are properly tested on the different supported platforms.

### Additional context

Before pushing this change, let's complete the basic e2e tests in #133

Angelmmiguel:
This change is blocked until [Wasmtime 10.0.0 is released](https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md#1000). I'm waiting for a fix regarding opening directories in Windows: https://github.com/bytecodealliance/wasmtime/pull/6348.

Angelmmiguel:
Wasmtime 10.0.0 was released a few hours ago! 🎉 

I'll open tomorrow a new PR to update to this version.

Angelmmiguel:
Upgrading OpenSSL to fix [RUSTSEC-2023-0044](https://rustsec.org/advisories/RUSTSEC-2023-0044)

PR #164, opened 2023-06-23
fix: install openssl@3 when building wws in actions

Angelmmiguel:
When releasing `wws` v1.3, the `artifacts` [action was failing for macOS](https://github.com/vmware-labs/wasm-workers-server/actions/runs/5353724410/jobs/9709914752). The issue was related to the `openssl` crate upgrade on #132.

This is the error message from the job:

```
  Could not find directory of OpenSSL installation, and this `-sys` crate cannot
  proceed without this knowledge. If OpenSSL is installed and this crate had
  trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
  compilation process.

  Make sure you also have the development packages of openssl installed.
  For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.

  If you're in a situation where you think the directory *should* be found
  automatically, please open a bug at https://github.com/sfackler/rust-openssl
  and include information about your system as well as this message.

  $HOST = x86_64-apple-darwin
  $TARGET = aarch64-apple-darwin
  openssl-sys = 0.9.90
``` 

Based on the `openssl` crate documentation, [you need to install OpenSSL to build it](https://docs.rs/openssl/latest/openssl/#automatic). I assume the action runner was installing and old version that it's no longer valid for the crate and we need to explicitly install the latest one.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -70,8 +70,8 @@ jobs:
       if: ${{ matrix.build == 'macos' && matrix.arch == 'aarch64' }}
       run: |
         brew install pkg-config-wrapper
-        brew fetch --force --bottle-tag=arm64_monterrey openssl@3
-        brew install $(brew --cache --bottle-tag=arm64_monterrey openssl@3)
+        brew fetch --force --bottle-tag=arm64_monterey openssl@3
+        brew install $(brew --cache --bottle-tag=arm64_monterey openssl@3)
     - name: Install deps (Linux)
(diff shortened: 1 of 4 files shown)


Angelmmiguel:
I removed the `artifacts` action condition to run it on pull requests (I added it for testing). Here you have a successful action that created all the artifacts: https://github.com/vmware-labs/wasm-workers-server/actions/runs/5411363945

ereslibre (approved):
LGTM, thanks @Angelmmiguel 👏

Issue #165, opened 2023-06-26
Send HTTP requests from inside the workers (fetch support)


Angelmmiguel:
### Is your feature request related to a problem? Please describe.

When I write a worker, I can only rely on the data that comes from a local file (mounted folders), environments variables or hardcoded in code. This reduces the kind of projects I can develop.

### Describe the solution you'd like

I would like to call an external HTTP API from inside a worker. Every language has its own way to do it, so I want to use those common patterns. For example, the `fetch` method is the most common API for calling an HTTP endpoint in JavaScript.

To follow the capability-based approach, I want to configure the HTTP endpoints a worker can call. I want to avoid a worker calling random HTTP endpoints. For this, I would add a new section in the TOML file called "features". This object includes one configuration per feature like: 

```toml
[features]
[features.fetch]
allowed_domains = ["https://example.com"]
force_ssl = true
```

### Host implementation

- [x] Define the bindings with WIT (#168)
- [x] Implement the bindings on the host side (#168)
- [x] Configure the workers so developers configure the allowed requests. By default, all requests will be rejected. (#174)

### Language implementation

I will start with JavaScript, but I will extend the support to other languages soon:

- [x] JavaScript (#169)
- [x] Rust (#171)
- [x] Go (#173)

#### Future implementations

- Python (#176)
- Ruby (#177)

Angelmmiguel:
Now that we have support for Go, Rust and JavaScript, I plan to split the remaining languages (Python and Ruby) into separate tasks. Note that languages that requires an external runtime follows a different lifecycle and we can publish them afterwards.

With the new versions, Python and Ruby workers will continue working but they won't be able to make HTTP requests yet. We will work on adding support to those soon 😄

PR #168, opened 2023-07-04
feat: add the HTTP bindings to perform HTTP requests from workers

Angelmmiguel:
Define the new bindings to perform HTTP request from inside the workers. The outcome of this PR is that workers can import a `send-http-request` function to send HTTP requests through the host. This features opens many new possibilities and it's a pretty standard API in any worker-based platform.

It refs #165 

## Interface definition

For now, we are not using the [latest WIT syntax](https://github.com/bytecodealliance/wit-bindgen#wit-as-an-idl) (including `package`). The `wit-bindgen` version that supports it only works with components. Wasm Workers Server doesn't support components yet, but we will switch to them in the future, while keeping compatibility with core components.

For this reason, we're pinning the `wit-bindgen` dependency to the [v2.0 version](https://github.com/bytecodealliance/wit-bindgen/commit/cb871cfa1ee460b51eb1d144b175b9aab9c50aba) on the client side (I'll send a separate PR with these changes).

On the host side, we're using the [`wit-bindgen-backport` fork](https://github.com/fermyon/wit-bindgen-backport) from the [Fermyon team](https://fermyon.com). This fork is compatible with the latest Wasmtime version and the `wit-bindgen` v2 version.

## Changes

- Add the `.wit` files that defines the bindings and types
- Add a new folder structure in the `worker` crate for future bindings
- Implement the bindings on Wasm Workers Server using `reqwest`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,8 +68,10 @@ exclude = [
 actix-web = "4"
 anyhow = "1.0.66"
 lazy_static = "1.4.0"
+reqwest = "0.11"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.85"
+tokio = { version = "1.28", features = ["rt", "rt-multi-thread", "sync"] }
 toml = "0.7.0"
 wws-config = { path = "./crates/config" }
 wws-runtimes = { path = "./crates/runtimes" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ lazy_static = "1.4.0"
 reqwest = "0.11"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.85"
-tokio = { version = "1.28", features = ["rt", "rt-multi-thread", "sync"] }
+tokio = "1.28"
 toml = "0.7.0"
 wws-config = { path = "./crates/config" }
 wws-runtimes = { path = "./crates/runtimes" }
--- a/crates/worker/src/bindings/http.rs
+++ b/crates/worker/src/bindings/http.rs
@@ -72,7 +72,7 @@ impl Http for HttpBindings {
 
                     builder = builder.body(body);
 
-                    let response = match builder.send().await {
+                    match builder.send().await {
                         Ok(res) => {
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -68,8 +68,10 @@ exclude = [
 actix-web = "4"
 anyhow = "1.0.66"
 lazy_static = "1.4.0"
+reqwest = "0.11"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.85"
+tokio = { version = "1.28", features = ["rt", "rt-multi-thread", "sync"] }
```
ereslibre, line 74:
```suggestion
tokio = "1.28"
```

This is no longer necessary given how we implemented the tokio executor on a separate thread.

crates/worker/src/bindings/http.rs
```diff
@@ -0,0 +1,116 @@
...
+wit_bindgen_wasmtime::export!({paths: ["../../wit/core/http.wit"]});
+use http::{Http, HttpError, HttpMethod, HttpRequest, HttpRequestError, HttpResponse};
+
+pub use http::add_to_linker;
+
+pub struct HttpBindings {}
+
+impl HttpBindings {
+    /// Map the reqwest error to a known http-error
+    fn map_reqwest_err(e: &reqwest::Error) -> HttpError {
```
ereslibre, line 17: This could be implemented as an `impl From<HttpBindings> for HttpError`, and then use `into()` instead of calling to `map_reqwest_err`.

This is more idiomatic I think.

ereslibre (commented):
LGTM! Thanks @Angelmmiguel. Some comments :)

crates/worker/src/bindings/http.rs
Angelmmiguel: +100! I'm gonna redo this. I though that I cannot implement a trait from a external struct, but it seems `From` / `Into` is an exception.

Cargo.toml
Angelmmiguel: Thank you! I already remove it.

ereslibre (approved):

PR #169, opened 2023-07-04
feat: add the fetch and more APIs to the JavaScript kit

Angelmmiguel:
_This PR is built on top of #168. We will need to merge and rebase this branch before_

The goal of this PR is to introduce more methods and common APIs to the JavaScript kit. It introduces the following APIs:

- `fetch`
- `console`
- `URLSearchParams`
- `TextEncoder` and `TextDecoder`
- A more complete `Response` implementation
- Better error management

It refs #165

## Building the shim

Before, the shim was a single `glue.js` file. All JavaScript code was hardcoded as we were not using any bundling system. To start using external libraries, I added [esbuild, a fast bundler for JavaScript](https://esbuild.github.io/). The shim code is available in the `kits/javascript/shims` folder and it contains all the files and references different external libraries.

As part of the build process of the JS engine, it builds the shim and injects it in the final Wasm module.

## External libraries

To avoid re-implementing all these APIs manually, we are using existing libraries: 

- [@sinonjs/text-encoding](https://www.npmjs.com/package/@sinonjs/text-encoding)
- [@ungap/url-search-params](https://www.npmjs.com/package/@ungap/url-search-params)
- [http-status](https://www.npmjs.com/package/http-status)


kits/javascript/src/error.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright 2023 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// List of runtime errors
+pub enum RuntimeError {
+    InvalidBinding(String),
+}
```
ereslibre, line 7:
```suggestion
    InvalidBinding(String),
```

Looking at this type it's not very clear to me what is the inner String. We can be more explicit and descriptive at the type-level by doing this:

```suggestion
pub enum RuntimeError {
    InvalidBinding { invalid_export: String },
}
```

ereslibre: This also allows us to grow the type information in a backwards and descriptive compatible way.

ereslibre (commented):
LGTM! A couple minor comments.

kits/javascript/src/error.rs
Angelmmiguel: Sounds good! I'm gonna apply it 😄

ereslibre (approved):

PR #174, opened 2023-07-11
feat: allow configuring the http_requests via TOML

Angelmmiguel:
In Wasm Workers Server, we want to follow the capability-based model that Wasm and WASI are enforcing. By default, workers cannot make HTTP requests. To enable it, you need to add a related configuration for the worker and enable only the specific requests you want.

For example:

```toml
name = "js-fetch"
version = "1"

[features]
[features.http_requests]
allowed_methods = ["GET"]
allowed_hosts = ["jsonplaceholder.typicode.com"]
allow_http = false
```

The default values for these config params are:

| Param | Default value |
| --- | --- |
| allowed_methods | `["GET", "POST", "PUT", "PATCH", "DELETE"]` |
| allowed_hosts | `[]` |
| allow_http | `false` | 

It refs #165

### DIFF
--- a/crates/worker/src/bindings/http.rs
+++ b/crates/worker/src/bindings/http.rs
@@ -16,6 +16,22 @@ pub struct HttpBindings {
     pub http_config: HttpRequestsConfig,
 }
 
+/// Implement the conversion between HttpMethod and
+/// http::Method
+impl From<HttpMethod> for reqwest::Method {
+    fn from(value: HttpMethod) -> Self {
+        match value {
+            HttpMethod::Get => Method::GET,
+            HttpMethod::Post => Method::POST,
+            HttpMethod::Put => Method::PUT,
+            HttpMethod::Patch => Method::PATCH,
+            HttpMethod::Delete => Method::DELETE,
+            HttpMethod::Options => Method::OPTIONS,
+            HttpMethod::Head => Method::HEAD,
+        }
+    }
+}
+
 /// Map the reqwest error to a known http-error
 /// HttpError comes from the HTTP bindings
 impl From<reqwest::Error> for HttpError {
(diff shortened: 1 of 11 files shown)


crates/worker/src/features/http_requests.rs
```diff
@@ -0,0 +1,33 @@
+// Copyright 2023 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::Deserialize;
+
+#[derive(Deserialize, Clone, Default)]
+pub struct HttpRequests {
```
ereslibre, line 7: Nit: What about `HttpRequestsConfig`?

crates/worker/src/features/http_requests.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub allowed_methods: Vec<String>,
+    /// Allow HTTP requests
+    #[serde(default = "default_false")]
+    pub allow_http: bool,
+}
+
+/// Turn the given configuration false by default
+fn default_false() -> bool {
+    false
+}
```
ereslibre, line 22:
```suggestion 
```

crates/worker/src/features/http_requests.rs
ereslibre, line 7: We can implement `impl Default for HttpRequests`, and let serde do the defaulting based on that, so there's no need to do per-field `serde(default)`.

```rust
#[derive(Deserialize, Clone)]
#[serde(default)]
pub struct HttpRequests {
}

impl Default for HttpRequests {
  fn default() -> HttpRequests {
    HttpRequests {
      // our defaults here...
    }
  }
}
```

ereslibre (commented):
This is great! Some comments to have more idiomatic code :)

crates/worker/src/features/http_requests.rs
Angelmmiguel: Sounds good! 👍

Angelmmiguel: Done!

ereslibre (approved):
Great! LGTM 👏

crates/worker/src/bindings/http.rs
```diff
@@ -39,8 +43,59 @@ impl Http for HttpBindings {
...
+        })?;
+        let method = match req.method {
+            HttpMethod::Get => Method::GET,
+            HttpMethod::Post => Method::POST,
+            HttpMethod::Put => Method::PUT,
+            HttpMethod::Patch => Method::PATCH,
+            HttpMethod::Delete => Method::DELETE,
+            HttpMethod::Options => Method::OPTIONS,
+            HttpMethod::Head => Method::HEAD,
+        };
```
ereslibre, line 60: We can probably handle this somewhere so that methods that need this conversion can get a little cleaner

Angelmmiguel: Good point. I'm gonna move it to a separate trait implementation.

PR #181, opened 2023-07-18
feat: add lib.rs with gated child crates

ereslibre:
This allows for Wasm Workers Consumers to depend on the toplevel WWS crate as a library, gating different features behind cargo features.

The cargo features that will permeate in the toplevel `lib.rs` depend on the usage that third parties will make of our API and I think is worth adding as needed.

This change keeps the ability for third parties to consume explicit subcrates if they want. However, if they consume more than one subcrate, when they bump the WWS version, they will need to do so in different places in their `Cargo.toml`. This toplevel `lib.rs` helps in this regard, by allowing a consumer to depend on the toplevel `lib.rs`, and only activate the features they need through Cargo features.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,12 @@ wws-project = { workspace = true }
 reqwest = { version = "0.11", features = ["blocking"] }
 
 [features]
+default = ["all"]
+all = ["wws_config", "wws_router", "wws_server"]
+wws_config = []
+wws_router = []
+wws_server = []
+
 vendored-openssl = ["wws-project/vendored-openssl"]
 
 [workspace]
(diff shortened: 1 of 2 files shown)


ereslibre:
Although I have not yet opened the PR for the containerd-wasm-shims project, the change as a consumer looks like this: https://github.com/deislabs/containerd-wasm-shims/compare/main...ereslibre:wws-toplevel-lib-rs

If this gets merged, that PR can be opened with minor modifications (e.g. the branch to pull wws from)

Angelmmiguel (approved):
LGTM! We can export more subcrates as required 👍

PR #182, opened 2023-07-19
feat: use custom error types in internal and external crates

ereslibre:
Builds on top of https://github.com/vmware-labs/wasm-workers-server/pull/181, although that PR is not strictly necessary for this one.

Fixes: https://github.com/vmware-labs/wasm-workers-server/issues/73

Opening it as a draft for now, when I'm happy with the final solution I'll make a final PR.

PR #185, opened 2023-07-20
feat: use more descriptive types for enabling features on serve

ereslibre:
This adds more descriptive types to enable/disable features such as the panel on the public API `serve()` function. With this change, consumers such as the `containerd-wasm-shims` will be more clear on the intent of their arguments.

This PR is on top of https://github.com/vmware-labs/wasm-workers-server/pull/182, but only the last commit matters. It will be rebased when https://github.com/vmware-labs/wasm-workers-server/pull/182 and https://github.com/vmware-labs/wasm-workers-server/pull/181 merge.


src/main.rs
```diff
@@ -196,7 +196,7 @@ async fn main() -> std::io::Result<()> {
             routes,
             &args.hostname,
             args.port,
-            args.enable_panel,
+            args.enable_panel.into(),
```
ereslibre, line 199: The conversion from `bool` to the custom type is done this way, so that there is no need for `crates/server` to know anything about clap, which is very specific to the CLI layer.

crates/router/src/lib.rs
```diff
@@ -54,6 +55,10 @@ impl Routes {
         Self { routes, prefix }
     }
 
+    pub fn iter(&self) -> impl Iterator<Item = &Route> {
+        self.routes.iter()
+    }
```
Angelmmiguel, line 60: 👏

crates/router/src/route.rs
```diff
@@ -4,15 +4,18 @@
...
     path::{Component, Path, PathBuf},
+    sync::RwLock,
 };
 use wws_config::Config as ProjectConfig;
 use wws_worker::Worker;
 
 lazy_static! {
     static ref PARAMETER_REGEX: Regex = Regex::new(r"\[\w+\]").unwrap();
     static ref DYNAMIC_ROUTE_REGEX: Regex = Regex::new(r".*\[\w+\].*").unwrap();
+    pub static ref WORKERS: RwLock<HashMap<String, Worker>> = RwLock::new(HashMap::default());
```
Angelmmiguel, line 18: This is a great optimization! We can point to a centralize location for workers' data. My only comment here is the API we're exposing to other crates.

Having to work with a `RwLock` directly seems fine for now, as we're the main consumers of these crates. However, I would use a higher level struct that interacts with this resource. 

What do you think?

crates/server/src/lib.rs
```diff
@@ -16,88 +16,140 @@ use actix_web::{
...
+    }
+}
 
 #[derive(Default)]
 pub(crate) struct DataConnectors {
     kv: KV,
 }
 
+#[derive(Clone)]
+pub struct ServeOptions {
```
Angelmmiguel, line 51: This is great! 👏

crates/server/src/lib.rs
```diff
@@ -16,88 +16,140 @@ use actix_web::{
...
+    pub root_path: PathBuf,
+    pub base_routes: Routes,
+    pub hostname: String,
+    pub port: u16,
+    pub panel: Panel,
+    pub stderr: Option<PathBuf>,
+    pub cors_origins: Option<Vec<String>>,
```
Angelmmiguel, line 58: I see this value is duplicated in `AppData`. Shall we keep it only in the there?

crates/server/src/lib.rs
```diff
@@ -16,88 +16,140 @@ use actix_web::{
...
-pub async fn serve(
-    root_path: &Path,
-    base_routes: Routes,
-    hostname: &str,
-    port: u16,
-    panel: bool,
-    stderr: Option<&Path>,
-    cors_origins: Option<Vec<String>>,
-) -> Result<Server> {
+pub async fn serve(serve_options: ServeOptions) -> Result<Server> {
```
Angelmmiguel, line 96: Some arguments are "mandatory" for the server initialization. I can think of `root_path`, `routes`, `hostname` and `port`. The other arguments are optional. What do you think about keeping them as independent arguments?

```suggestion
pub async fn serve(root: &Path, routes: Routes, hostname: &str, port: u16, options: ServeOptions) -> Result<Server> {
```

I'm thinking on other similar APIs, which sets mandatory arguments and then, an open `options` object for extra params.

Angelmmiguel (commented):
LGTM in general! I'm glad to see we're providing a better interface to set the different options! Added some comments 😄

crates/router/src/route.rs
ereslibre: Yes, I think that would be a great approach. Should we tackle that on a separate PR, or take advantage of it and add it here?

crates/server/src/lib.rs
ereslibre: It's probably a bit repetitive with AppData; we can probably put the common elements in a dedicated struct and use it in both places.

crates/router/src/route.rs
Angelmmiguel: If you can add it to this PR, that would be great. For me, it's also fine if you do it in a separate PR 😄

crates/server/src/lib.rs
ereslibre: Yes, I commented on https://github.com/vmware-labs/wasm-workers-server/pull/185/files/64a2bec46185777c075b06b31f6d9489e2e2187e#r1308965573: basically the idea is that the user builds a ServeOptions and it's transformed into an AppData. However, I think the common attributes might better be in a shared struct that both point to. I will update the PR.

crates/server/src/lib.rs
ereslibre: Updated the PR with all optional attributes. I prefer the struct option so that arguments are 'named', as opposed to positional heterogeneous arguments on the caller.

crates/server/src/lib.rs
ereslibre: This structure is for providing arguments to the `serve` call, whereas `AppData` is for workers to read/write "runtime" information.

The flow in this case is the following: CLI/third party consumer -> `ServeOptions` -> `AppData` <-> workers.

Angelmmiguel: Ahh I see. So it makes sense to keep that info in both places 👍

crates/server/src/lib.rs
Angelmmiguel: Sounds good! I'm not very familiar with Rust conventions yet, so either options work for me! 👍 to merge it as a struct.

Angelmmiguel (approved):
LGTM! 👏

PR #186, opened 2023-07-20
chore: place wws under /opt/wws and keep /app in prebuilt images too

ereslibre:
Place wws under /opt in the container images that use prebuilt binaries, in the same place as the container images that build the project to reduce the drift and surprises depending if a user build the container on their own machine, or if they pull `ghcr.io/vmware-labs/wws:preview` for example.

### DIFF
--- a/image/Prebuilt.dockerfile
+++ b/image/Prebuilt.dockerfile
@@ -3,7 +3,8 @@
 # GitHub actions
 
 # Retrieve the certificates to install runtimes later on.
-FROM --platform=$TARGETPLATFORM bitnami/minideb:latest AS certs
+FROM --platform=$TARGETPLATFORM bitnami/minideb:latest AS sysroot
+RUN mkdir -p /target/app /target/opt
 RUN install_packages ca-certificates
 
 # Build the final image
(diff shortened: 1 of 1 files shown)


Angelmmiguel (commented):
Thanks for the update @ereslibre! Totally agree on consolidating the `wws` binary location in the different images. 

I added a comment regarding the `ENTRYPOINT` after this change.

image/Prebuilt.dockerfile
```diff
@@ -14,9 +15,10 @@ LABEL org.opencontainers.image.source=https://github.com/vmware-labs/wasm-worker
...
 LABEL org.opencontainers.image.licenses="Apache-2.0"
 
-COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
-COPY --chmod=755 ./wws-$TARGETARCH /wws
+COPY --from=sysroot /target/app /app
+COPY --from=sysroot /target/opt /opt
+COPY --from=sysroot /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --chmod=755 ./wws-$TARGETARCH /opt/wws
 
 ENTRYPOINT ["/wws"]
```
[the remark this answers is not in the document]
ereslibre: Thanks for the heads up! My bad after rebasing on top of https://github.com/vmware-labs/wasm-workers-server/pull/187

image/Prebuilt.dockerfile
```diff
@@ -3,7 +3,8 @@
 # GitHub actions
 
 # Retrieve the certificates to install runtimes later on.
-FROM --platform=$TARGETPLATFORM bitnami/minideb:latest AS certs
+FROM --platform=$TARGETPLATFORM bitnami/minideb:latest AS sysroot
+RUN mkdir -p /target/app /target/opt
```
Angelmmiguel, line 7: Are these folders required on the `sysroot` container? I see we're copying files later but I believe those folders are empty.

ereslibre: It's just a trick because we cannot rely on `mkdir` on the resulting image, so we just `COPY` the folders from the `sysroot` repository, that are created on this layer

Angelmmiguel: I see! Good point 😄

Angelmmiguel (approved):
LGTM! 👏

Issue #166, opened 2023-06-30
Simplify how you run wws commands in the container


Angelmmiguel:
### Is your feature request related to a problem? Please describe.

The `wws` container runs the `wws` CLI with a set of default flags. To configure this behavior, the container uses the CMD directive:

https://github.com/vmware-labs/wasm-workers-server/blob/93e116584df1c580aa8e071076062dd059a88f76/image/Prebuilt.dockerfile#L20

This solution works well when you don't want to modify or add argument. However, if you want to add a flag or run a project from a remote repository, the following command doesn't work:

```
docker run -p 8080:8080 ghcr.io/vmware-labs/wws:v1.3.0 \
  https://github.com/vmware-labs/wasm-workers-server.git \
    --git-folder "examples/js-basic" \
    -i --enable-panel --host 0.0.0.0
```

The CMD directive gets fully override and it fails because "https://github.com/vmware-labs/wasm-workers-server.git" is not a valid binary. You need to refer to the `wws` binary directly, which it's not in the `PATH` making difficult to add / remove parameters.

The following code works, but it's not trivial:

```
docker run -p 8080:8080 ghcr.io/vmware-labs/wws:v1.3.0 \
  /wws https://github.com/vmware-labs/wasm-workers-server.git \
    --git-folder "examples/js-basic" \
    -i --enable-panel --host 0.0.0.0
```

### Describe the solution you'd like

I want the container to run "like a binary". The following command should work out of the box:

```
docker run -p 8080:8080 ghcr.io/vmware-labs/wws:v1.3.0 \
  https://github.com/vmware-labs/wasm-workers-server.git \
    --git-folder "examples/js-basic" \
    -i --enable-panel --host 0.0.0.0
```

These are the requirements:

* The container always run the `wws` CLI. For that, we can use the [ENTRYPOINT](https://docs.docker.com/engine/reference/builder/#entrypoint) directive which fits perfectly for this use case.
* Keep the `CMD` directive, but only for the default arguments (`/app` and `--host 0.0.0.0`). You can override the default values by adding arguments as showed in the previous example. [There is a great example in the Docker documentation](https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example).

For convenience, the `wws` should be present in the `PATH` in case you need to update the default entrypoint. Pointing it to the binary location (`/wws`) is not trivial.

dierbei:
I want to try to solve this problem.

PR #187, opened 2023-07-25
fix: update Dockerfile

dierbei:
I increased the `ENTRYPOINT` in the Dockerfile to default to the path to the wws executable.

And the `CMD` command can override the default value when running the image.

Fixes: #166

### DIFF
--- a/image/Dockerfile
+++ b/image/Dockerfile
@@ -31,5 +31,5 @@ RUN apt-get update && \
 RUN mkdir -p /app
 RUN mkdir -p /opt
 COPY --from=build-wws ${WWS_BUILD_DIR}/build/wws /opt
-
-CMD ["/opt/wws", "/app/", "--host", "0.0.0.0"]
+ENTRYPOINT ["/opt/wws"]
+CMD ["/app/", "--host", "0.0.0.0"]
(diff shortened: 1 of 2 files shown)


image/Dockerfile
```diff
@@ -31,5 +31,5 @@ RUN apt-get update && \
 RUN mkdir -p /app
 RUN mkdir -p /opt
 COPY --from=build-wws ${WWS_BUILD_DIR}/build/wws /opt
-
-CMD ["/opt/wws", "/app/", "--host", "0.0.0.0"]
+ENTRYPOINT ["/opt/wss"]
```
ereslibre, line 34: Typo :)

```suggestion
ENTRYPOINT ["/opt/wws"]
```

image/Dockerfile
ereslibre, line 1: Could you please also update https://github.com/vmware-labs/wasm-workers-server/blob/e7c22c6af59e285f97dfd344e24002d1dab35c48/image/Prebuilt.dockerfile to follow the same approach?

ereslibre (changes requested):
Thank you for your contribution @dierbei! A couple of comments before we can merge it :)

image/Dockerfile
dierbei: thanks for your advice.

image/Dockerfile
dierbei: ok.

PR #193, opened 2023-08-10
panel forward

### LINKED ISSUE #192: When panel is enabled, only `/index.html` can be directly visited
### Describe the bug

When the panel is enabled `--enable-panel`, the logic that handles the panel endpoints will only serve `/index.html` and any assets that are discovered under `client/dist` when building.

The logic that handles this is at: https://github.com/vmware-labs/wasm-workers-server/blob/549387b83be50b1b2ec7e04782c629a2c2e38965/crates/panel/src/handlers/panel.rs#L18-L29

There are other virtual paths such as `/_panel/workers` that are not covered by any of this cases.

Ideally, to avoid duplicating routing in multiple places, this logic should:

1. If path is empty, default to `index.html`, as per the current logic.
2. If an asset exists, serve it, as per the current logic.
3. Otherwise, serve a static page, that will perform an AJAX request to the current path and will forward the response to the client.

This way, we keep the current Single Page Application (SPA) approach, while making arbitrary paths directly visitable (e.g. reload endpoint, write path in location bar and hit enter...)

### Reproduction steps

Run wws with `--enable-panel`. Visit `/_panel/workers`. The server will return a 404  HTTP status code.

### Expected behavior

Visiting `/_panel/workers` directly should be equivalent to visiting `/` and navigating through the UI to the list of workers.

### Additional context

_No response_

dierbei:
Fix #192

### DIFF
--- a/crates/panel/src/handlers/panel.rs
+++ b/crates/panel/src/handlers/panel.rs
@@ -12,19 +12,30 @@ use rust_embed::RustEmbed;
 #[folder = "client/dist/"]
 struct Asset;
 
-/// Find the static assets of the administration panel
 #[actix_web::get("/_panel{_:.*}")]
 pub async fn handle_static_panel(path: web::Path<String>) -> impl Responder {
-    let path = if path.len() == 0 {
+    let path = if path.is_empty() {
         "index.html"
     } else {
         path.as_str().strip_prefix('/').unwrap()
     };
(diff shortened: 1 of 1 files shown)


Angelmmiguel (approved):
LGTM! Thank you very much for your contribution @dierbei 😄

Issue #213, opened 2023-09-05
Document catch all routes


ereslibre:
### Is your feature request related to a problem? Please describe.

https://github.com/vmware-labs/wasm-workers-server/pull/197 implemented catch all routes.

This is an issue to track writing the documentation accompanying this feature. The process of expanding documentation is written down here: https://github.com/vmware-labs/wasm-workers-server/tree/main/docs.

### Describe the solution you'd like

Feature is documented :)

PR #214, opened 2023-09-12
Document catch all routes

ereslibre:
Fixes: https://github.com/vmware-labs/wasm-workers-server/issues/213


docs/docs/features/dynamic-routes.md
```diff
@@ -56,13 +56,61 @@ $ tree .
...
+│   └── [...slug]
+│       └── index.js
+└── [...slug]
+    └── index.js
+```
+
+Here, we have the same structure as in the previous example, but we have two catch-all, under two different roots:
+
+- One catch-all, `[...slug]/index.js` is serving all requests, except for requests whose path starts with `/about` or `/other`.
+- Another catch-all, `other/[...slug]/index.js` serves all requests under the `/other` path.
```
Angelmmiguel, line 101: This is great @ereslibre ! I would add some examples in which the segments are defined in the file directly (`[...slug].js`).

docs/docs/features/dynamic-routes.md
```diff
@@ -56,13 +56,61 @@ $ tree .
...
+
+### Routing priority
+
+Given catch-all routes could potentially shadow other routes, it is important to settle precedence when routing requests. The rule of thumb is more specific routes win. This is, a route with no catch-all will always against a route with catch-all when they are at the same depth.
```
Angelmmiguel, line 105: Minor suggestions:

```suggestion
Given catch-all routes could potentially shadow other routes, it is important to settle precedence when routing requests. **The rule of thumb is more specific routes win**. For example, a route with no catch-all will always against a route with catch-all when they are at the same depth.
```

examples/js-catchall/other/[...slug]/index.js
```diff
@@ -0,0 +1,58 @@
+/**
```
Comment (Angelmmiguel) [inline examples/js-catchall/other/[...slug]/index.js:1]: I would use a here a file directly: `examples/js-catchall/other/[...slug].js`

Angelmmiguel (approved):
LGTM in general! Added some minor comments, but approving it in advance so you can merge after fixing it :).

Thank you for fixing all the tables btw!

ereslibre:
> Thank you for fixing all the tables btw!

Haha, my editor does automatically when I edit a table :D

examples/js-catchall/other/[...slug]/index.js
Comment (ereslibre) [reply examples/js-catchall/other/[...slug]/index.js:1]: I can add the one you propose too, but thought that having two catch-all at the directory level to show precedence at that level is also nice and somewhat documenting.

I will add this one too!

Comment (ereslibre) [reply examples/js-catchall/other/[...slug]/index.js]: Ah nevermind I see what you meant here, you are right. Updated.

PR #217, opened 2023-09-13
feat: add Zig mount example

ereslibre:


kits/zig/worker/src/worker.zig
```diff
@@ -90,7 +97,7 @@ pub const Output = struct {
         }
 
         // prepare writer for json
-        var out_buf: [1024]u8 = undefined;
+        var out_buf: [4096]u8 = undefined;
```
ereslibre, line 100: When this merges I'm going to open an issue to need an allocator here instead of fixed sizes on the stack.

Angelmmiguel: Sounds good 👍

Angelmmiguel (approved):
LGTM!

PR #224, opened 2023-09-27
feat: allow to preload ML models when running inference

### LINKED ISSUE #215: Add named models support on AI workers

Currently, workers that run inference must retrieve the ML model in memory to pass it back to the host. This limits the number of models we can run as Wasm has a memory limit of 4Gb (32-bit).

The WASI-NN proposal recently introduced the ["Named Models" feature](https://github.com/WebAssembly/wasi-nn/issues/36). This feature allows the host to preload a set of models and expose them to the modules using a label. The model no longer need to load the model in memory. It only needs to reference the model using the right label.

`wws` can take advantage of this feature by adding new configuration parameters to the `feature.wasi_nn` object. Then, `wws` will preload those models and expose them to the workers.

The `wasmtime-wasi-nn` and `wasi-nn` ([see example](https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasi-nn/examples/classification-example-named)) crates already supports this feature.

_No response_

### Additional context

_No response_

Angelmmiguel:
Introduce the ML model preloading by leveraging the "[Named models](https://github.com/WebAssembly/wasi-nn/issues/36)" feature from WASI-NN. Now, workers don't need to load the model files and send it to the host, but loading it in the host directly. I added a new example that uses this feature and produces the same result as the [rust-wasi-nn example](https://github.com/vmware-labs/wasm-workers-server/tree/main/examples/rust-wasi-nn).

I also introduced some new configuration properties to the `wasi_nn` feature:

```
name = "wasi-nn-mobilenet"
version = "1"

[[features.wasi_nn.preload_models]]
backend = "openvino"
provider = { type = "local", dir = "./_models/mobilenet/" }
```

The `preload_models` array allows to configure multiple models. The `provider` configures the way `wws` retrieves the models. Currently, only the local provider is available but we plan to add more.

## Limitations

Even though we're preloading the models on the host, this process still happens at request level. Based on some comments, the [Wasmtime WASI-NN library is not protected against concurrent access](https://github.com/bytecodealliance/wasmtime/blob/5272e3abc99c3f1e7829e9fd2c9bf2b7fac17639/src/commands/run.rs#L793). To avoid multiple workers accessing the same context, I'll keep this at request level for now.

In the future, my goal is to load models once at worker initialization.

It closes #215


crates/worker/src/features/wasi_nn.rs
```diff
@@ -2,12 +2,100 @@
...
+        match self {
+            Self::None => write!(f, "none"),
+            Self::Openvino => write!(f, "openvino"),
+        }
+    }
+}
+
+/// Available providers to load Wasi NN models.
+#[derive(Deserialize, Clone, Debug)]
+#[serde(rename_all = "lowercase", tag = "type")]
```
ereslibre, line 42: Nice :)

ereslibre (approved):
Super!, LGTM :)

PR #225, opened 2023-09-29
feat: add support for loading WebAssembly components

ereslibre:
This adds support for loading WebAssembly modules and components. It also adds components under `examples` and adds them to the e2e test suite.

Missing bits:

- Register wasi-nn for components
- Migrate http wit and register for components


Angelmmiguel:
Related to #227

crates/runtimes/src/modules/external.rs
```diff
@@ -90,14 +90,28 @@ impl Runtime for ExternalRuntime {
...
+                    )?
+                    .args(&self.metadata.args)
+                    .map_err(|_| errors::RuntimeError::WasiContextError)?;
+            }
+            CtxBuilder::Preview2(ref mut builder) => {
+                builder
+                    .preopened_dir(
+                        Dir::open_ambient_dir(&self.store.folder, ambient_authority())?,
+                        preview2::DirPerms::all(),
+                        preview2::FilePerms::all(),
```
Angelmmiguel, line 109: This is cool! We can add more granularity to the worker configuration file in the future.

crates/runtimes/src/runtime.rs
```diff
@@ -3,7 +3,12 @@
 
 use crate::errors::Result;
 
-use wasmtime_wasi::WasiCtxBuilder;
+use wasmtime_wasi::{preview2, WasiCtxBuilder};
+
+pub enum CtxBuilder {
+    Preview1(WasiCtxBuilder),
+    Preview2(preview2::WasiCtxBuilder),
+}
```
Angelmmiguel, line 11: Great solution!

crates/worker/src/lib.rs
```diff
@@ -99,122 +233,185 @@ impl Worker {
...
+                ModuleOrComponent::Component(component) => {
+                    let (command, _instance) = preview2::command::Command::instantiate_async(
+                        &mut store,
+                        component,
+                        &component_linker,
+                    )
+                    .await
+                    .unwrap();
+                    let _ = command
+                        .wasi_cli_run()
```
Angelmmiguel, line 395: So, the `run` method in the CLI mode is the equivalent of the `_start` method from WASI preview1, isn't it? I'm trying to gather all the pieces you had to work on to make this transition 😄

Angelmmiguel (commented):
This is a huuuuge effort @ereslibre ! Congratulations for such a complete PR. I'm really happy that `wws` starts supporting components and regular core modules. This PR will open a new set of possibilities for the project 😄.

I added some minor comments, but everything LGTM!

crates/worker/src/errors.rs
```diff
@@ -7,13 +7,15 @@ pub type Result<T> = std::result::Result<T, WorkerError>;
 
 #[derive(Debug)]
 pub enum WorkerError {
-    BadWasmModule,
+    BadWasmModuleOrComponent,
```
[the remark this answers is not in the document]
ereslibre: The problem is that we cannot be certain about the user intent. Right now, the logic looks like:

```mermaid
flowchart TD
  A[Is this a module?];
  A -- Yes --> B[Load];
  A -- No --> C[Is this a component?];
  C -- Yes --> B[Load];
  C -- No ---> D[Error];
```

I think we can do a bit better: we can check the magic number, removing the trial and error. The problem with the current approach is that we cannot be explicit about the error, because the final error in the current flow is always about a component failed to load (because is the last check we do).

With the magic number check I think we can provide the user valuable information in the case their core module or component could not be loaded, and get rid of the trial and error for module and component.

crates/worker/src/lib.rs
ereslibre: Yes, exactly. It is very clear from the adapter code:

https://github.com/bytecodealliance/wasmtime/blob/270e92225da3acb5b0a583546b2d4598cb59ae2a/crates/wasi-preview1-component-adapter/src/lib.rs#L61-L70

ereslibre: (Note: this applies to command executable model, the reactor model has a slightly different adapter)

crates/worker/src/errors.rs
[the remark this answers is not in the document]
ereslibre: I'll update the PR with this :)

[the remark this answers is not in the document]
Angelmmiguel: Ahh I see. Thanks for the explanation! Yeah, checking the magic number would be ideal to avoid a try-error approach.

crates/worker/src/lib.rs
Angelmmiguel: Thank you 👍

crates/worker/src/errors.rs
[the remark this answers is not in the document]
ereslibre: Addressed in https://github.com/vmware-labs/wasm-workers-server/pull/225/commits/7e24b21b6474ebc938bfb5f825875419aa4d0a10

crates/worker/src/stdio.rs
```diff
@@ -1,6 +1,9 @@
 use std::io::Cursor;
 use wasi_common::pipe::{ReadPipe, WritePipe};
-use wasmtime_wasi::WasiCtxBuilder;
+use wasmtime_wasi::preview2;
+use wws_runtimes::CtxBuilder;
+
+const MAX_OUTPUT_BYTES: usize = 10240;
```
Angelmmiguel, line 6: Please, could you add a comment about why setting this specific limit?

examples/Makefile
```diff
@@ -1,38 +1,31 @@
-.PHONY: all rust-* pdf-create
+SUBDIRS = rust-basic rust-fetch rust-kv rust-wasi-nn rust-wasi-nn-preload rust-pdf-create
+COMPONENTS = components/rust-basic components/rust-kv
```
Angelmmiguel, line 2: Nice improvement 😄

Angelmmiguel (approved):
LGTM! Just added a minor comment 😄

crates/worker/src/lib.rs
```diff
@@ -70,26 +120,123 @@ impl Worker {
...
+            Ok(ModuleOrComponent::Component(
+                Component::from_binary(&engine, &bytes).map_err(|err| {
+                    errors::WorkerError::BadWasmComponent {
+                        error: format!("{:?}", err),
+                    }
+                })?,
+            ))
+        } else {
+            Err(errors::WorkerError::BadWasmCoreModuleOrComponent)
+        }?;
```
Angelmmiguel, line 152: This is amazing!

crates/worker/src/stdio.rs
ereslibre: Right, and this is not nice...

ereslibre: Good catch, thank you :) Updated :)

ereslibre:
We can merge on green then, thanks for the reviews and feedback @Angelmmiguel!

Issue #175, opened 2023-07-11
Load workers dynamically


Angelmmiguel:
### Is your feature request related to a problem? Please describe.

When you start Wasm Workers Server, it retrieves all the workers in the given location (remote Git repo or local folder). After this initialization, it cannot add new workers on the fly and it requires you to restart the entire server.

Since we have an administration panel (#167), it would be nice to allow creating new workers on the fly or even edit existing ones. This is also related to the ability of "watching" running workers and reloading them (#157) 

### Describe the solution you'd like

I would like Wasm Workers Server to load new workers dynamically. Instead of a static list of `Routes` and a fixed set of URLs, the `wasm_handler` should receive any request that was not managed by other handler. Then, it iterates over the `Routes` and find the one that can reply it. The `Routes` struct must be mutable, so we can add new workers / routes under demand.

That approach opens new uses cases like the ones mentioned above.

Angelmmiguel:
After starting working on this issue, I noticed some challenges here. Currently, we're registering specific routes for every worker and use the `handle_worker` method to reply. For example, we associate `/demo` from `demo.js` to `handle_worker`. However, we cannot follow this approach anymore as we may add workers lately.

Then, I added a `default_service` that captures any non-processed requests and pass them to `handle_worker`. The default service caused collisions with `actix-files`, which it's the library we're using to serve the [static files](https://workers.wasmlabs.dev/docs/features/static-assets/). By default, `wws` mounts the assets in `/` which takes precedence and tries to reply to all requests.

For this reason, I had to invert the approach. Now, I plan to register all the static files in specific routes and then, use the `handler_worker` as the `default_service`. It works, although it's a bit tedious as know we need to manually iterate over the `public` folder.

All this complexity comes from the fact that we're also supporting "pretty" routes in the `public` folder. Like serving `/public/index.html` in `/`. If we don't support it, we can just associate any path that doesn't include a `.` character to the `handle_worker` method. Other option could be to mount the `public` folder in the `/public` prefix, although some specific static assets like `robots.txt` must be at the root level.

What do you think @ereslibre?

ereslibre:
> However, we cannot follow this approach anymore as we may add workers lately.

Why we cannot continue following this approach? Routes would be the discovered ones from the filesystem, plus the ones registered through the API. We might error out if an API call tries to register a path that clashes with an existing one, be it a discovered or a previously registered path.

[...]

PR #226, opened 2023-10-02
feat: load static assets manually so all other requests rely on wasm_handler

Angelmmiguel:
This is the first PR to complete #175. Before this PR, the default service that captured any non-processed request was the `actix-files` service ([via `Files`](https://docs.rs/actix-files/0.6.2/actix_files/struct.Files.html)). Since we wanted to mount static files into `/`, `Files` was capturing al `/.*` routes. For workers, `wws` was mounting their routes explicitly. 

In #175, our intention is the opposite. We want `workers_handler` to manage any unknown route, so we can add new routes when the server is running. My first approach was to find a way to pass a request to `worker_handler` from `Files` using the `default_handler`. The `Files` service triggers this handler when it doesn't find any matching file. However, the types didn't match and it was an extra step before processing the request.

For this reason, I changed the approach and set all the static assets routes manually. Then, `wws` sends any unknown route to the `wasm_handler` to detect if there's any worker that manages it. In fact, this is a more correct approach as static files have a higher priority than workers (like dynamic).

### DIFF
--- a/crates/server/src/handlers/assets.rs
+++ b/crates/server/src/handlers/assets.rs
@@ -11,7 +11,7 @@ use std::io::{Error, ErrorKind};
 /// like "public/{uri}/index.html" and "public/{uri}.html".
 ///
 /// If no file is present, it will try to get a default "public/404.html"
-pub async fn handle_assets(req: &HttpRequest) -> Result<NamedFile, Error> {
+pub async fn handle_assets(req: HttpRequest) -> Result<NamedFile, Error> {
     let root_path = &req
         .app_data::<Data<AppData>>()
         .expect("error fetching app data")
@@ -25,7 +25,7 @@ pub async fn handle_assets(req: &HttpRequest) -> Result<NamedFile, Error> {
     // Same as before, but the file is located at ./about.html
     let html_ext_path = root_path.join(format!("public{uri_path}.html"));
 
-    if file_path.exists() {
+    if file_path.exists() && !uri_path.is_empty() && uri_path != "/" {
         NamedFile::open_async(file_path).await
     } else if uri_path.ends_with('/') && index_folder_path.exists() {
         NamedFile::open_async(index_folder_path).await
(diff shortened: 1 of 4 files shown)


crates/server/src/handlers/assets.rs
```diff
@@ -25,7 +25,7 @@ pub async fn handle_assets(req: &HttpRequest) -> Result<NamedFile, Error> {
     // Same as before, but the file is located at ./about.html
     let html_ext_path = root_path.join(format!("public{uri_path}.html"));
 
-    if file_path.exists() {
+    if file_path.exists() && !uri_path.is_empty() && uri_path != "/" {
```
Angelmmiguel, line 28: We need this new condition to avoid returning the `public` folder information.

crates/server/src/handlers/worker.rs
```diff
@@ -42,16 +42,8 @@ pub async fn handle_worker(req: HttpRequest, body: Bytes) -> HttpResponse {
...
     let selected_route = app_data.routes.retrieve_best_route(req.path());
-    if let Some(route) = selected_route {
-        // First, check if there's an existing static file. Static assets have more priority
-        // than dynamic routes. However, I cannot set the static assets as the first service
-        // as it's captures everything.
-        if route.is_dynamic() {
-            if let Ok(existing_file) = handle_assets(&req).await {
-                return existing_file.into_response(&req);
-            }
-        }
```
Angelmmiguel, line 53: Since we're explicitly mounting the static files in certain routes, this is not required anymore.

crates/server/src/lib.rs
```diff
@@ -75,11 +74,38 @@ impl From<ServeOptions> for AppData {
...
+
+    // Pre-create the KV namespaces
+    let data_connectors = Data::new(RwLock::new(data));
+
+    // Static assets
+    let mut static_assets =
+        StaticAssets::new(&serve_options.root_path, &serve_options.base_routes.prefix);
+    static_assets
+        .load()
+        .expect("Error loading the static assets");
```
Angelmmiguel, line 106: I extracted all the initialization befure the `HttpServer::new` call. Actix runs this block on every thread, so we avoid calculating static assets multiple times.

Angelmmiguel:
@ereslibre don't review this yet, as I need to rethink about this issue. I may close this PR and open a more well-defined issue with a proper set of goals and definition

Angelmmiguel:
For now, I'm going to close this PR!

PR #228, opened 2023-10-04
chore: avoid using format when building asset paths and double-check path

Angelmmiguel:
The `handle_assets` handler relies on `uri.path()`, which it's already sanitized by `actix`. We introduced a new set of checks to confirm the filepath is valid before trying to read it, to avoid future issues.

### DIFF
--- a/crates/server/src/handlers/assets.rs
+++ b/crates/server/src/handlers/assets.rs
@@ -1,10 +1,17 @@
-// Copyright 2022 VMware, Inc.
+// Copyright 2022-2023 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::AppData;
 use actix_files::NamedFile;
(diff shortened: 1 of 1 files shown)


crates/server/src/handlers/assets.rs
```diff
@@ -1,10 +1,17 @@
...
 use crate::AppData;
 use actix_files::NamedFile;
 use actix_web::{web::Data, HttpRequest};
 use std::io::{Error, ErrorKind};
+use std::path::{Component, Path};
+
+/// Checks if the given component is not normal or absolute. If the component
+/// is not valid, it returns true.
+fn component_is_valid(component: Component) -> bool {
+    matches!(component, Component::Normal(_) | Component::RootDir)
```
ereslibre, line 13: matches! is cool :)

ereslibre (approved):
LGTM 👏

PR #229, opened 2023-10-04
feat: manage static assets properly in Unix and Windows

Angelmmiguel:
Since URIs use `/` as the separator, `wws` didn't return static assets in a subfolder in Windows. I took the opportunity to refactor part of the code and add tests to ensure `wws` calculates asset paths properly. For that, I split the process into different steps:

* Retrieve URI
* Clean up the path
* Retrieve the path in the `public` folder that points to a possible static asset

I included a battery of tests for those methods and I'm planning to add some e2e tests for static assets.


crates/server/src/handlers/assets.rs
```diff
@@ -4,7 +4,45 @@
...
+/// in the public folder, only "normal" components are valid.
+fn clean_up_path(uri: &str) -> PathBuf {
+    // First split the URI as it always uses the /.
+    let path = PathBuf::from_iter(uri.split('/'));
+
+    let valid_components: Vec<Component<'_>> = path
+        .components()
+        // Keep only normal components
+        .filter(|c| matches!(c, Component::Normal(_)))
+        .collect();
```
ereslibre, line 22: I cannot help but think that this is somehow weird because we are just filtering out non-normal components. If actix didn't perform clean up, then a resource of the form `/some/path/../../logo.png` will become `/some/path/logo.png`. If this asset happens to exist in the filesystem we will serve it, but this might be something different than what the user was expecting.

ereslibre (commented):
Thanks a lot for taking care of this and for adding a nice amount of tests.

Just a comment, I know actix filters this for us now, but I wonder about the logic itself, if we cannot rely on that filtering.

crates/server/src/handlers/assets.rs
Angelmmiguel: At this point, the relative components have been solved by actix. So, URI shouldn't include any `..` or `.` values. For this reason, I'm applying this final double check when building the file path. Just in case some weird encoding passes the actix filtering and can be interpreted by Rust as a parent component (`..`). 

Note that this is not the final path, but the portion that will be appended to the `public_folder` later on in the `retrieve_asset_path` method. In that method I also check the final path is a subpath of the public folder. 

Maybe the naming is not the most appropriate 😅.

ereslibre (approved):
LGTM 👍

Issue #230, opened 2023-10-05
wasi-nn is not registered on the linker


ereslibre:
### Describe the bug

wasi-nn is not registered on the linker, neither for core modules, or for components. This makes it impossible to use the inference features of wws.

### Reproduction steps

1. Install the latest version of wws
2. Setup the examples/rust-wasi-nn-preload (read its README.md)
3. Execute wws pointing to this example
4. The inference is not working

### Expected behavior

The inference works.

PR #231, opened 2023-10-05
chore: register wasi-nn on the linker

ereslibre:
Fixes: https://github.com/vmware-labs/wasm-workers-server/issues/230


Angelmmiguel (approved):
LGTM! :+1:

PR #235, opened 2023-10-16
fix: manage Headers instances in the JS kit

### LINKED ISSUE #233: Error when initializing Headers in the JavaScript kit
### Describe the bug

The JavaScript kit allows you to define the response headers. You can pass a second argument to the `Response` type that may include a `headers` key with the response headers. When I pass a `Headers` object to the the `Response`, the server returns an error because the response is not properly formatted:

```
let headers = new Headers({});
return new Response("Hello World", { headers });
```

When running this example, I get the following error:

```
<p>There was an error running this function</p>
```

The output from the worker is the following:

```
{"data":"Hello World","headers":{"headers":{}},"status":200,"kv":{}}
```

### Reproduction steps

1. Create a new `index.js` file with the following content:

```js
const reply = (_request) => {
  // Return the response
  let headers = new Headers({});
  return new Response("Hello World", { headers });
}

// Subscribe to the Fetch event
addEventListener("fetch", event => {
  return event.respondWith(reply(event.request));
});
```

2. Run `wws`
3. Try to get the response from the worker

### Expected behavior

[When the `Response` type receives a `Headers` instance](https://github.com/vmware-labs/wasm-workers-server/blob/2b72b055db0f16dd40173f1eb0e99d798e5b8c2b/kits/javascript/shims/types/response.js#L13), it should reuse it instead of trying to build a new `Headers` instance. This is a potential implementation:

```js
if (options.headers instanceof Headers) {
  this.headers = options.headers;
} else {
  this.headers = new Headers(options.headers || {});
}
```

I can also check if `options.headers` is an Object before calling the `new Headers` method with `options.headers`.

### Additional context

_No response_

Angelmmiguel:
Properly manage `Headers` instances by reusing them when a `Response` receives it. Before, a `Headers` object was causing a `{ headers: { headers: {} }` object, which was wrong. I also took the opportunity to improve the Headers parsing by only allowing valid types (`Object` or `Headers`), and valid header values (`string`).

It fixes #233


ereslibre (approved):
Thanks for this fix :)

PR #236, opened 2023-10-16
feat: add support for ECMAScript modules in the JavaScript kit

### LINKED ISSUE #149: Support modules in JavaScript workers

At a first approach, the JavaScript workers follow the Web Workers API. It requires to subscribe to a specific event and then, process the request and provide a response:

https://github.com/vmware-labs/wasm-workers-server/blob/51da71082e4cd41d787b88354d1e14c69ce22f43/examples/js-params/fixed.js#L27-L30

However, the current JavaScript SDK emulates this behavior, so it doesn't provide any Web API behind it. In addition to that, other platforms like Cloudflare and Vercel started to simplify this syntax in favor of ECMA modules:

```javascript
export default {
  async fetch(request) {
    const body = `Hello world!`;
    return new Response(body);
  },
};
```

This syntax is simpler and more convenient.

I want to write workers using the ECMA modules approach, while keeping compatibility with existing modules based on the Web Workers API.

It requires to change the current JavaScript SDK to:

* [ ] Manage ECMA modules
* [ ] Detect the current approach
* [ ] Run the worker to get the response

_No response_

### Additional context

_No response_

Angelmmiguel:
Thanks to the new Javy version, we can now eval JavaScript modules with the `eval_module` method. To use this new method, [I had to update the `javy` version to 2.0.0](https://github.com/bytecodealliance/javy/blob/main/crates/quickjs-wasm-rs/CHANGELOG.md#200---2023-08-17). The upgrade caused some errors in the build process due to crate incompatible versions (`wasmtime-wasi` requires `bytes = "^1.4"` while `javy` requires `bytes = "=1.1.0"`. Since I consider these two projects as separate entities, I just excluded it from the workspace.

Finally, I added a new `js-modules` example that showcases it.

It closes #149


kits/javascript/src/main.rs
```diff
@@ -49,7 +49,19 @@ fn main() {
         },
     }
 
-    context.eval_global("handler.js", &contents).unwrap();
+    // Checks if the given code uses ECMAScript modules.
+    if contents.contains("export default") {
```
ereslibre, line 53: I guess this is fine for our purposes now, but this would identify a comment "// export default" as so, even if it wasn't really JS code.

ereslibre (approved):
LGTM as a first approach, I think we can improve later.

kits/javascript/src/main.rs
Angelmmiguel: I also added a third regex to detect exported objects like:

```js
export default {
  async fetch() {
    // ...
  }
}
```

ereslibre:
LGTM modulo the windows CI

PR #249, opened 2023-10-20
fix: add missing console.trace method

### LINKED ISSUE #248: console.trace is missing in the JS kit
### Describe the bug

The `console.trace` method is used by frameworks like HonoJS to provide extra information when capturing errors. This method is not present in the current SDK and is causing the framework to fail.

### Reproduction steps

1. Clone the project and go to the `examples/js-hono` folder
2. Edit the `src/index.js` file and add a `throw new Error("error!"`)` to any handler
3. Run the example and access to the handler

### Expected behavior

Instead of a generic error by `wws`, Hono should manage it and provide the error. Currently, it's failing because it tries to call `console.trace`.

### Additional context

_No response_

Angelmmiguel:
Add the missing `console.trace` method to the JavaScript SDK.

It closes #248

### DIFF
--- a/kits/javascript/shims/bindings.js
+++ b/kits/javascript/shims/bindings.js
@@ -53,6 +53,9 @@ import { Response } from "./types/response";
     },
     warn(msg) {
       this.log(msg);
+    },
+    trace(msg) {
+      this.log(msg);
     }
   }
 


ereslibre (approved):
Great, thanks! 👏

PR #255, opened 2023-11-24
feat: start using rustls instead of openssl

### LINKED ISSUE #137: Switch to rustls to avoid compiling OpenSSL

Currently, the project relies on OpenSSL to perform `https` requests to pull the different runtimes. We are using the `reqwest` crate for it.  When compiling the project to Linux with the `musl` toolchain, we started facing errors [due to missing dependencies](https://github.com/vmware-labs/wasm-workers-server/pull/77#issue-1558066793).

To fix it, we configured the `openssl` dependency to build the project using the `vendored` feature. We only applied this to the `musl` targets:

https://github.com/vmware-labs/wasm-workers-server/blob/feedff9668bd35309033f7b484d2777c1820f088/Cargo.toml#L39-L43

We want to avoid compiling this dependency as it causes other issues in the past.

The goal is to use the [rustls](https://github.com/rustls/rustls) crate. We can configure reqwest to use it via features. However, we cannot do it yet due to an issue with the Windows aarch64 target in the `ring` dependency. The problem is fix, although the new release is not public yet. 

You can track the current status on: https://github.com/briansmith/ring/issues/1551

_No response_

### Additional context

_No response_

Angelmmiguel:
Before, `wws` was using OpenSSL to establish external connections using the `reqwest` crate. Initially, we wanted to use `rustls`, but the compilation was failing for Windows arm64. After closing all the tasks described in #137, we can now switch to `rustls`.

In the past, we had multiple issues and CVEs due to the OpenSSL dependency. In addition to that, we had to compile it from scratch on every build so it was taking extra time. Also, we had to include `ca-certificates` in the containers images as we are using `scratch`. From now on, we will rely on `rustls` related certs (webpki).

It closes #137

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ wws-server = { workspace = true }
 wws-project = { workspace = true }
 
 [dev-dependencies]
-reqwest = { version = "0.11", features = ["blocking"] }
+reqwest = { version = "0.11", features = ["rustls", "blocking"] }
 
 [features]
 default = ["all"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,6 @@ wws_config = []
 wws_router = []
 wws_server = []
 
-vendored-openssl = ["wws-project/vendored-openssl"]
-
 [workspace]
 members = [
   "crates/api-manage",
@@ -77,7 +75,7 @@ exclude = [
 [workspace.dependencies]
 actix-web = "4"
 lazy_static = "1.4.0"
-reqwest = "0.11"
+reqwest = { version = "0.11", features = ["rustls"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.85"
 tokio = "1.28"
@@ -99,4 +97,3 @@ wasmtime-wasi = "13.0.0"
 wasmtime-wasi-nn = "13.0.0"
 wasi-common = "13.0.0"
 path-slash = "0.2.1"
-openssl = { version = "=0.10.55" }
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -34,7 +34,7 @@ wws-server = { workspace = true }
 wws-project = { workspace = true }
 
 [dev-dependencies]
-reqwest = { version = "0.11", features = ["blocking"] }
+reqwest = { version = "0.11", features = ["rustls", "blocking"] }
```
ereslibre, line 37:
```suggestion
reqwest = { version = "0.11", features = ["rustls-tls", "blocking"] }
```

As per https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L41

ereslibre: We might want to use `rustls-tls-webpki-roots` as well, so that we can provide a static binary that contains the root certificates bundled in the binary itself, without the need for the host system running wws to have ca-certificates or similar installed.

This has its problems of course, as for example, root certificate revocation, but I think it should be fine for now.

ereslibre (commented):
Super! Thanks @Angelmmiguel! Just a couple of comments :), LGTM :)

Cargo.toml
Angelmmiguel: Good catch! Unfortunately, I'm going to close this PR due to #256. We still need to find an alternative that works for git2-rs :/

Angelmmiguel:
Closing for now as this feature is blocked by #256

Cargo.toml
ereslibre: Ouch, right.

PR #270, opened 2024-04-16
docs: Add examples for all languages into the documentation

hwclass:
Screenshot 2024-04-17 at 08 49 43


docs/package-lock.json
ereslibre, line 1: I think we can avoid adding the package lock; can you please add it to the .gitignore file?

ereslibre (commented):
Thanks a lot for your contribution @hwclass! It LGTM 👏, just one comment regarding the lock file and I think we can merge it, once the CLA process has been completed. Once merged I'll take care of publishing the docs :)

docs/package-lock.json
hwclass: Hi @ereslibre 

Thanks for your message. I think we need to keep track of the metadata of the packages installed as how it has been mentioned in npm's docs here: https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json#description

Once you give the thumbsup, we're good to go, I think 👍

ereslibre: In this case it shouldn't be relevant, because we are using docusaurus to generate the documentation, and that dependency is pinned to a specific version, so the package.lock generates more noise than it solves in this case, IMO.

Before we can move forward we have to wait for the CLA to go through, thanks again for your contribution! 👍

hwclass: hey @ereslibre, ok then, just pushed the updated commit. Thanks for your comments 🙌

hwclass: Hey @ereslibre 

It is already 3 weeks, is there any progress on the CLA or there is sth. else that I need to do to move forward with the PR?

fyi. @flavio @nebhale

ereslibre: Hello @hwclass. First of all, thank you for your contribution and I'm sorry it's taking this long for the CLA to be sorted out.

Let me find out and I'll come back to you as soon as I know what's the reason for the delay.

ereslibre: Hello @hwclass,

Sorry for the long time to reply and thank you for your contribution. After some long discussions, this project is currently in the process of being archived. However, the development of this project will continue as an independent community fork present at https://github.com/webassemblylabs/wasm-workers-server.

We would be grateful if you could recreate your contribution under that fork and we will work on applying it.

